PR #36, opened 2017-10-25
Fix the "Create Process"-Button

### LINKED ISSUE #17: Button 'Create Process' does not work
The 'Create Process' Button does not work.

Pushing the button calls:
/#/start/cc8c62d0-051e-4245-a3eb-02c3f2b9e94a

The uuid is hard coded in the charon code.
The uuid does not belong to the 'Prozess erstellen' process.

Maybe hard coding the id is not the best way?
But if it is, I think it should be something like:
/#/processdef/<PROCESS_UUID>/start

Regards

Vyperus:
## What did you change?

Closes #17 

Try to create a new Process via the create button, you will be redirected to enter the initial data for the new process.

## PR-Checklist


src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,25 +1,32 @@
...
-  private createProcess: string = environment.createProcess;
   private subscriptions: Array<Subscription>;
 
-  constructor(eventAggregator: EventAggregator, consumerClient: ConsumerClient) {
+  constructor(eventAggregator: EventAggregator, consumerClient: ConsumerClient, router: Router) {
     this.eventAggregator = eventAggregator;
     this.consumerClient = consumerClient;
+    this.router = router;
+
+    this.consumerClient.once('renderUserTask', (userTaskConfig: IUserTaskEntity) => {
```
heikomat: i'm not sure this would behave correctly every time. have you tested the button after jumping back and forth between parts of the app a couple of times, and trying the button multiple times without reloading the website?

Paulomart: This would be triggered by other userTasks comming in as well, any process would trigger a navigation. From UX this would be very strage to the user.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -86,4 +93,9 @@ export class ProcessDefList {
     }
     return this._processes.data;
   }
+
+  public createProcess(): void {
+    this.consumerClient.startProcessByKey('CreateProcessDef');
```
Paulomart: Instead I would save the started process instance id there and **then** wait for a dynamic ui request on that and only that process instance.

PR #48, opened 2017-10-28
✨ Process Def List is now sorted

### LINKED ISSUE #7: Sort for all lists
Any list that is display should either have a fixed sort order or an option to set sort order.

Vyperus:
## What did you change?

This PR adds a basic sorting method for the processdef-list.
This Sorting method sorts lexicographical for the key.

Closes #7 

Save an already saved process in the Detail-View. The Process will now stay in the same place as before when reloading the List-View

## PR-Checklist

### DIFF
--- a/src/modules/processdef-list/processdef-list.ts
+++ b/src/modules/processdef-list/processdef-list.ts
@@ -73,7 +73,14 @@ export class ProcessDefList {
     if (this._processes === undefined) {
       return [];
     }
-    return this._processes.data;
+    return this._processes.data.sort((a: IProcessDefEntity, b: IProcessDefEntity) => {
+      if (a.key < b.key) {
+        return -1;
+      } else if (a.key > b.key) {
+        return 1;
+      }
+      return 0;
+    });
   }
 
   public async createProcess(): Promise<void> {


ppoetschke:
I don‘t think it‘s a good idea to sort client side as it requires to load all entities. There are options to do correct sorting - even together with pagination - in the backend.

Vyperus:
I agree with you! But then we need to relocate the mentioned Issue. :)

ppoetschke:
Not really because sort and query options should be implemented client side and passed through to the backend. They are already implemented in the backend.

heikomat:
@ppoetschke even so this would still be more of a consumerClient issue, and less of a charon issue. The consumerClient needs to give charon the option to get a sorted list.

ppoetschke:
Instead of only using limit and offset as params for `getProcessDefList`in `consumer_client` the attributes of `IPublicQueryOptions` of `core_contracts` should be implemented. IPublicQueryOptions already supports sorting on the backend correctly.
The orderBy option should be set by Charon and passed through by consumer_client. I consider setting a static orderBy option in consumer_client as anti pattern as it prevents sorting initiated by the user i.e. by clicking on a column header.
Please add appropriate issues in the involved repos if necessary.

PR #167, opened 2018-02-19
🐛 :bug: Fix setFirstElement function

alexanderkasten:
## What did you change?

This PR selects now always an element (after import or first attach of detailview).
- first try to set the startevent
- then try to set another element
- if that fails, it selects the process 

- import a diagram with a startevent and other elements => startevent is selected
- import a diagram without a startevent => another element is selected
- import a diagram without any elements => process is selected

## PR-Checklist

### DIFF
--- a/src/modules/property-panel/property-panel.ts
+++ b/src/modules/property-panel/property-panel.ts
@@ -91,6 +91,8 @@ export class PropertyPanel {
         if (!startEvent && process.flowElements) {
           startEvent = process.flowElements[0];
         }
+      } else if (process.laneSets && process.laneSets[0].lanes) {
+        startEvent = process.laneSets[0].lanes[0];
       }
 
       if (!startEvent) {
--- a/src/modules/property-panel/property-panel.ts
+++ b/src/modules/property-panel/property-panel.ts
@@ -39,7 +39,8 @@ export class PropertyPanel {
       this.extensionsIndextab,
     ];
 
-    this.checkForEachIndexTabisSuitable();
+    this.updateIndexTabsSuitability();
+    this.checkIndexTabSuitability();
 
     this.eventBus.on(['element.click', 'shape.changed', 'selection.changed'], (event: IEvent) => {
 
@@ -58,7 +59,8 @@ export class PropertyPanel {
         this.elementInPanel = event.newSelection[0];
       }
 
-      this.checkForEachIndexTabisSuitable();
+      this.updateIndexTabsSuitability();
+      this.checkIndexTabSuitability();
     });
 
     this.setFirstElement();
@@ -119,13 +121,20 @@ export class PropertyPanel {
     }
   }
 
-  private checkForEachIndexTabisSuitable(): void {
-    this.indextabs.forEach((indextab: IIndextab) => {
+  private updateIndexTabsSuitability(): void {
+    for (const indextab of this.indextabs) {
       indextab.canHandleElement = indextab.isSuitableForElement(this.elementInPanel);
-      if (indextab.title === this.currentIndextabTitle && !indextab.canHandleElement) {
-        this.currentIndextabTitle = this.generalIndextab.title;
-      }
+    }
+  }
+
+  private checkIndexTabSuitability(): void {
+    const currentIndexTab: IIndextab = this.indextabs.find((indextab: IIndextab) => {
+      return indextab.title === this.currentIndextabTitle;
     });
+
+    if (!currentIndexTab.canHandleElement) {
+      this.currentIndextabTitle = this.generalIndextab.title;
+    }
   }
 
 }
--- a/src/modules/property-panel/property-panel.ts
+++ b/src/modules/property-panel/property-panel.ts
@@ -102,8 +102,14 @@ export class PropertyPanel {
   }
 
   private processHasLanes(process: IModdleElement): boolean {
-    return  (process.laneSets !== undefined || process.laneSets !== null) &&
-            (process.laneSets[0].lanes !== undefined || process.laneSets[0].lanes !== null);
+    const processHasLaneSets: boolean = process.laneSets !== undefined && process.laneSets !== null;
+    if (!processHasLaneSets) {
+      return false;
+    }
+
+    const processHasLanes: boolean = process.laneSets[0].lanes !== undefined && process.laneSets[0].lanes !== null;
+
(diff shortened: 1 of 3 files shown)


src/modules/property-panel/property-panel.ts
```diff
@@ -77,15 +77,26 @@ export class PropertyPanel {
...
         return element.$type === 'bpmn:Process';
       });
-      const startEvent: IModdleElement = process.flowElements.find((element: any ) => {
-        return element.$type === 'bpmn:StartEvent';
-      });
-      if (startEvent.$type !== 'bpmn:StartEvent') {
-        startEvent.id = process.flowElements[0].id;
+
+      if (process.flowElements) {
+        startEvent = process.flowElements.find((element: any ) => {
```
Vyperus: Don't use any

src/modules/property-panel/property-panel.ts
```diff
@@ -77,15 +77,26 @@ export class PropertyPanel {
...
+
+        if (!startEvent && process.flowElements) {
+          startEvent = process.flowElements[0];
+        }
+      } else if (process.laneSets && process.laneSets[0].lanes) {
+        startEvent = process.laneSets[0].lanes[0];
+      }
+
+      if (!startEvent) {
+        startEvent = process;
```
Vyperus: It is probably better to call it "startElement" not "startEvent" since it is not always the start event.

src/modules/property-panel/property-panel.ts
```diff
@@ -80,19 +80,30 @@ export class PropertyPanel {
...
       const process: IModdleElement = definitions.rootElements.find((element: IModdleElement) => {
         return element.$type === 'bpmn:Process';
       });
-      const startEvent: IModdleElement = process.flowElements.find((element: any ) => {
-        return element.$type === 'bpmn:StartEvent';
-      });
-      if (startEvent.$type !== 'bpmn:StartEvent') {
-        startEvent.id = process.flowElements[0].id;
+
+      if (process.flowElements) {
```
Vyperus: Move this to it's own variable like this:
`const processHasFlowElements: boolean = process.flowElements !== undefined && process.flowElements !== null`

then use this variable: 
`if(processHasFlowElements) {...}`

src/modules/property-panel/property-panel.ts
```diff
@@ -80,19 +80,30 @@ export class PropertyPanel {
...
+        firstElement = process.flowElements.find((element: IModdleElement ) => {
+          return element.$type === 'bpmn:StartEvent';
+        });
+
+        if (!firstElement && process.flowElements) {
```
Vyperus: You already checked "process.flowElements" already, you don't need to check it again here.

src/modules/property-panel/property-panel.ts
```diff
@@ -80,19 +80,30 @@ export class PropertyPanel {
...
+          firstElement = process.flowElements[0];
+        }
+      } else if (process.laneSets && process.laneSets[0].lanes) {
```
Vyperus: Check "process.laneSets" like this: 

```
private processHasLanes(process): boolean {
  return process.laneSets !== undefined && process.laneSets !== null && ...
}

Comment (SteffenKn) [reply src/modules/property-panel/property-panel.ts]: Why should we use a function for this and a variable for the previous `if`?

File: src/modules/property-panel/property-panel.ts [hunk 6]
```diff
@@ -80,29 +70,56 @@ export class PropertyPanel {
...
       const elementRegistry: IElementRegistry = this.modeler.get('elementRegistry');
-      const elementInPanel: IShape = elementRegistry.get(startEvent.id);
+      const elementInPanel: IShape = elementRegistry.get(firstElement.id);
 
       this.modeler.get('selection').select(elementInPanel);
     }));
   }
 
+  private processHasLanes(process: IModdleElement): boolean {
+    return  (process.laneSets !== undefined || process.laneSets !== null) &&
```
Vyperus: This cannot work! Also you can divide this into two variables to increase the readability. Do it like this:

```typescript
private processHasLanes(process: IModdleElement): boolean {
  const processHasLaneSets: boolean = process.laneSets !== undefined && process.laneSets !== null;
  if(!processHasLaneSets) {
    return false;
  }

  const processHasLanes: boolean = process.laneSets[0].lanes !== undefined && process.laneSets[0].lanes !== null;

  return processHasLanes;
}
```

src/modules/property-panel/property-panel.ts
```diff
@@ -80,29 +70,56 @@ export class PropertyPanel {
...
+  }
+
   private xmlChanged(newValue: string, oldValue: string): void {
     if (oldValue) {
       this.setFirstElement();
       this.updateIndextab(this.generalIndextab);
     }
   }
 
+  private checkForEachIndexTabisSuitable(): void {
```
Vyperus: Please refactor this into two methods:

```typescript
private updateIndexTabsSuitability(): void {
  this.indextabs.forEach((indextab: IIndextab) => {
    indextab.canHandleElement = indextab.isSuitableForElement(this.elementInPanel);
  });
}
```

and

```typescript
private checkIndexTabSuitability(): void {
  const currentIndexTab: IIndexTab = this.indextabs.find((indextab: IIndextab) => {
    return indextab.title === this.currentIndextabTitle;
  });

  if (!currentIndexTab.canHandleElement) {
    this.currentIndexTabTitle = this.generalIndexTab.title;
  }
}
```

alexanderkasten: why? where is one of these functions ever used separately? there`s no use case for split the function in two functions

5ebastianMeier: Readability is also a valid use case and this improves it. Also @Vyperus version doesn't check for the need to update the index tab title inside the loop which is good, because it is meant to be updated only once, not as often as the loop is executed.

heikomat: I like this suggestion! we could improve on it a tiny bit more, by replacing the `forEach` with a `for/of`-loop. Although in this case this is mostly for readability (i think `for/of` are even better to read than `forEach`)

```TypeScript
private updateIndexTabsSuitability(): void {
  for (const indextab of this.indextabs) {
    indextab.canHandleElement = indextab.isSuitableForElement(this.elementInPanel);
  };
}
```

PR #221, opened 2018-03-19
✨ Automatic detection of free port

heikomat:
## What did you change?

**SEE [THIS DIFF](https://github.com/process-engine/bpmn-studio/pull/221/files?w=1) FOR A BETTER OVERVIEW!**
(in this diff, the lines where only the indentation changed are not shown)

I made the electron-version of the bpmn-studio automatically choose a free port to run the backend on, if 8000 is already in use.

This however overrides the "last used server"-functionality we currently have, because it works by setting the new backend-location as last used server before the bpm-studio starts.

I also added the current baseRoute as absolute-positioned label to the bottom-left of the application
bildschirmfoto 2018-03-19 um 18 10 29

Something like a server-history would be nice to improve this situation.

A lot of stuff in the electron.js was just inendet, because it needs to happen after the port was found to not create a race-condition. To see the changed files without the indention-changes, klick here:

https://github.com/process-engine/bpmn-studio/pull/221/files?w=1

- Start something on port 8000
- Start the electron-version of the bpmn-studio
- See how it still works
- Click the gear-icon and see how the server-port is not 8000 but something random

## PR-Checklist

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -28,5 +28,5 @@
   </nav>
 
   <router-view></router-view>
-  <span class="small baseroute-text">${environment.bpmnStudioClient.baseRoute}</span>
+  <span class="small baseroute-text">Connected to: ${environment.bpmnStudioClient.baseRoute}</span>
 </template>
--- a/src/app.html
+++ b/src/app.html
@@ -28,4 +28,5 @@
   </nav>
 
   <router-view></router-view>
+  <span class="small baseroute-text">${environment.bpmnStudioClient.baseRoute}</span>
 </template>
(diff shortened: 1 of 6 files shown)


src/app.html
```diff
@@ -28,4 +28,5 @@
   </nav>
 
   <router-view></router-view>
+  <span class="small baseroute-text">${environment.bpmnStudioClient.baseRoute}</span>
```
Vyperus: I like this one, please add some context for the user. Just displaying the baseroute of the PE could lead to confusion.

PR #241, opened 2018-04-10
✨ Add File Association for the Electron Application

### LINKED ISSUE #247: File Association in die Electron App einbauen
#### Describe your issue

Die Electron App soll in der Lage sein BPMN Diagramme bzw. XMLs per Doppelklick auf das gewünschte Diagramm in die Datenbank aufzunehmen und das Diagramm im Modeler anzuzeigen.

## Aktueller Stand
OSX: Aktuell ist es möglich eine .bpmn Datei per Doppelklick in die Datenbank zu laden und sie im Modeler anzuzeigen, wenn BPMN-Studio noch nicht gestartet ist. Es ist **NICHT** möglich ein Diagramm zu importieren, wenn ein Fenster des BPMN-Studios geöffnet ist, oder kein Fenster, aber dafür aber eine aktive Instanz der Electron App (Icon im Dock) läuft.

Windows: Das verhalten von BPMN-Studio mit der File Association (und die generelle Funktionstüchtigkeit bezüglich des Backends) muss auf Windows getestet werden. 
https://github.com/process-engine/bpmn-studio/issues/253

---------
#241 

#### Issue checklist

MiguelACarballo:
## What did you change?

This PR adds file association for files with a `.bpmn` extension.

Currently it only opens the bpmn-studio app and 
safes the file path and file content into an object when double clicking on a `.bpmn` file.

Closes #247 

the https://github.com/process-engine/bpmn-studio/pull/243 pull request should merged before this, because this PR includes the changes from #243 

Bezüglich des genutzten NotificationServices ist hier noch ein relevantes Issue:
https://github.com/process-engine/bpmn-studio/issues/255

Info: Das process_engine_http package muss im `skeleton/process-engine-server` in der package.json auf `^1.1.0` stehen.

- Checkout the branch
- `npm install`
- `npm run build`
- `npm run electron-build`
- Install andstart BPMN-Studio with the generated .dmg file 
- Close it and double click on a `bpmn` file

The BPMN-Studio app should start now.

## PR-Checklist

### DIFF
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -123,29 +123,29 @@ getPort({port: 8000, host: '0.0.0.0'})
       app.quit();
     }
   });
-  app.on('will-finish-launching', function() {
-    let filePath;
+  let filePath;
 
-    electron.ipcMain.on('get_opened_file', (event) => {
-      if (filePath === undefined) {
-        event.returnValue = null;
-        return;
-      }
-
-      event.returnValue = {
-        path: filePath,
-        content: fs.readFileSync(filePath, 'utf8'),
-      }
-    });
-  
+  app.on('will-finish-launching', () => {
     // for windows
     if (process.platform == 'win32' && process.argv.length >= 2) {
       filePath = process.argv[1];
     }
     
     // for non-windows
-    app.on('open-file', function(event, path) {
+    app.on('open-file', (event, path) => {
       filePath = path;
     });
+  });
+
+  electron.ipcMain.on('get_opened_file', (event) => {
+    if (filePath === undefined) {
+      event.returnValue = {};
+      return;
+    }
+
+    event.returnValue = {
+      path: filePath,
+      content: fs.readFileSync(filePath, 'utf8'),
+    }
   });
 });
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -47,8 +47,6 @@ getPort({port: 8000, host: '0.0.0.0'})
       mainWindow = null;
     });
 
-    mainWindow.webContents.openDevTools();
-
     autoUpdater.checkForUpdates();
 
     autoUpdater.addListener('error', (error) => {
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -41,7 +41,7 @@ getPort({port: 8000, host: '0.0.0.0'})
       minHeight: 800,
       icon: path.join(__dirname, '../build/icon.png'),  // only for windows and linux
     });
-
+    
     mainWindow.loadURL(`file://${__dirname}/../index.html`);
     mainWindow.on('closed', () => {
       mainWindow = null;
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -41,7 +41,7 @@ getPort({port: 8000, host: '0.0.0.0'})
       minHeight: 800,
       icon: path.join(__dirname, '../build/icon.png'),  // only for windows and linux
     });
-    
+
     mainWindow.loadURL(`file://${__dirname}/../index.html`);
     mainWindow.on('closed', () => {
       mainWindow = null;
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -6,6 +6,7 @@ const notifier = require('electron-notifications');
 const isDev = require('electron-is-dev');
 const getPort = require('get-port');
 const fs = require('fs');
+const prereleaseRegex = /\d+\.\d+\.\d+-pre-b\d+/;
 
 if (!isDev) {
   const userDataFolder = process.env.APPDATA || (process.platform == 'darwin' ? process.env.HOME + '/Library/Preferences' : '/var/local');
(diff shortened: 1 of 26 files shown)


src/main.ts
```diff
@@ -10,6 +10,7 @@ export function configure(aurelia: Aurelia): void {
   if ((<any> window).nodeRequire) {
     const ipcRenderer: any = (<any> window).nodeRequire('electron').ipcRenderer;
     const newHost: string = ipcRenderer.sendSync('get_host');
+    const fileInfo: any = ipcRenderer.sendSync('get_opened_file');
```
LeoTT: I would rather use no type than `any` here.
TypeScript has type inference, which means if the right side has a type, the variable declaration will automatically use that type (unless you're overwriting it with a more general type like `any`). In this case the right side may have no typing, but you gain nothing from using `any` here.

That being said, you're not using `fileInfo` anywhere, so why declare a variable at all?

electron_app/electron.js
```diff
@@ -122,4 +123,29 @@ getPort({port: 8000, host: '0.0.0.0'})
       app.quit();
     }
   });
+  app.on('will-finish-launching', function() {
```
LeoTT: Could use arrow function here.

This event is only called once, right? Seems weird to set up eventlisteners within the callback of an eventlistener (because if this outer eventlistener gets triggered more than once you will end up with more callbacks per event than anticipated). Is there a reason not to set up those listeners on the same level?

electron_app/electron.js
```diff
@@ -122,4 +123,29 @@ getPort({port: 8000, host: '0.0.0.0'})
...
+    let filePath;
+
+    electron.ipcMain.on('get_opened_file', (event) => {
+      if (filePath === undefined) {
+        event.returnValue = null;
```
LeoTT: Is `null` necessary? I really don't like `null`. [here's why](https://github.com/5minds/javascript-guidelines/issues/8)
Issue is not closed but I have not heard any arguments or voices in favor of `null`.

electron_app/electron.js
```diff
@@ -122,4 +123,29 @@ getPort({port: 8000, host: '0.0.0.0'})
...
+      }
+    });
+  
+    // for windows
+    if (process.platform == 'win32' && process.argv.length >= 2) {
+      filePath = process.argv[1];
+    }
+    
+    // for non-windows
+    app.on('open-file', function(event, path) {
```
LeoTT: Is this explicitly for `non-windows`? Because it looks like that eventlistener is also set for windows.
Also, you could use an arrow function here.

heikomat: yes, it is also set on windows, but it's never fired there. We expected this event to work the same on every device, but it doesn't

src/modules/processengine/processengine.service.ts
```diff
@@ -2,13 +2,18 @@ import {IProcessDefEntity, IUserTaskEntity} from '@process-engine/process_engine
 import {inject} from 'aurelia-framework';
 import {IPagination, IProcessEngineRepository, IProcessEngineService, IProcessEntity} from '../../contracts';
 
-@inject('ProcessEngineRepository')
+@inject('ProcessEngineRepository', 'FileContent')
 export class ProcessEngineService implements IProcessEngineService {
 
   private repository: IProcessEngineRepository;
+  private fileInfo: any = undefined;
```
LeoTT: `any` type does nothing helpful here. Would be nice to add typing manually here if you know what fileinfo looks like.
maybe:
```javascript
{
    content: string,
}
```
Initializing with `undefined` happens automatically, so  you can delete `= undefined`.

src/modules/notification/notification.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  private _showNotification(notification: INotification): void {
+    if (notification.type === NotificationType.SUCCESS) {
+      this.toastrInstance.success(notification.message);
+    } else if (notification.type === NotificationType.ERROR) {
+      this.toastrInstance.error(notification.message);
+    } else if (notification.type === NotificationType.INFO) {
+      this.toastrInstance.info(notification.message);
+    } else if (notification.type === NotificationType.WARNING) {
+      this.toastrInstance.warning(notification.message);
+    }
```
LeoTT: You can make this more readable with a switchcase.

src/modules/processengine/processengine.service.ts
```diff
@@ -1,14 +1,31 @@
...
-import {IPagination, IProcessEngineRepository, IProcessEngineService, IProcessEntity} from '../../contracts';
+import {IFileInfo, IPagination, IProcessEngineRepository, IProcessEngineService, IProcessEntity, NotificationType} from '../../contracts/index';
+import {NotificationService} from './../notification/notification.service';
 
-@inject('ProcessEngineRepository')
+@inject('ProcessEngineRepository', 'FileContent', 'NotificationService')
 export class ProcessEngineService implements IProcessEngineService {
 
   private repository: IProcessEngineRepository;
+  private fileInfo: IFileInfo = undefined;
```
LeoTT: No need to initialize with undefined, happens by default.

src/modules/notification/notification.service.ts
```diff
@@ -0,0 +1,50 @@
+import {INotification, NotificationType} from '../../contracts/index';
+
+export class NotificationService {
+
+  private toastrInstance: any;
```
LeoTT: I would recommend to use `any` only for objects that could literally be anything (instead of something which has a definite type which you don't know).
You gain no information by using `any` and it overwrites type inference.
Moreover now when there's an `any` I don't know whether it stands just for a missing typing or it really means that it could be anything.

No typing is better than `any`-type for those cases imo.

PR #243, opened 2018-04-13
✨Feature/add importprocess button

### LINKED ISSUE #159: Import-Process-Button in processdef-list-view
#### Describe your issue

An "Import-Process"-Button needs to be implemented into the "list"-view.
This will make it much easier for the user to import processes and just start working on them.

#### Issue checklist

alexanderkasten:
## What did you change?

This PR adds an import BPMN button to the processdef-list.

Closes #159 

- checkout the branch
- `au run`
- import a BPMN file 

## PR-Checklist

### DIFF
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -40,7 +40,7 @@ getPort({port: 8000, host: '0.0.0.0'})
       minHeight: 800,
       icon: path.join(__dirname, '../build/icon.png'),  // only for windows and linux
     });
-    
+
     mainWindow.loadURL(`file://${__dirname}/../index.html`);
     mainWindow.on('closed', () => {
       mainWindow = null;
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -42,7 +42,6 @@ getPort({port: 8000, host: '0.0.0.0'})
     });
     
     mainWindow.loadURL(`file://${__dirname}/../index.html`);
-    // mainWindow.webContents.openDevTools();
     mainWindow.on('closed', () => {
       mainWindow = null;
     });
--- a/electron_app/electron.js
+++ b/electron_app/electron.js
@@ -40,8 +40,9 @@ getPort({port: 8000, host: '0.0.0.0'})
       minHeight: 800,
       icon: path.join(__dirname, '../build/icon.png'),  // only for windows and linux
     });
-
+    
     mainWindow.loadURL(`file://${__dirname}/../index.html`);
+    // mainWindow.webContents.openDevTools();
     mainWindow.on('closed', () => {
       mainWindow = null;
     });
--- a/src/modules/processdef-list/processdef-list.ts
+++ b/src/modules/processdef-list/processdef-list.ts
@@ -1,12 +1,13 @@
 import {BpmnStudioClient, IPagination, IProcessDefEntity, IUserTaskConfig} from '@process-engine/bpmn-studio_client';
 import {EventAggregator, Subscription} from 'aurelia-event-aggregator';
-import {inject, observable} from 'aurelia-framework';
+import {bindable, inject, observable} from 'aurelia-framework';
 import {Router} from 'aurelia-router';
-import {AuthenticationStateEvent} from '../../contracts/index';
+import {AuthenticationStateEvent, IProcessEngineService} from '../../contracts/index';
 import environment from '../../environment';
 
-@inject(EventAggregator, 'BpmnStudioClient', Router)
+@inject(EventAggregator, 'BpmnStudioClient', Router, 'ProcessEngineService')
 export class ProcessDefList {
+  private processEngineService: IProcessEngineService;
   private bpmnStudioClient: BpmnStudioClient;
   private eventAggregator: EventAggregator;
   private router: Router;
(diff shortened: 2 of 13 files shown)


src/modules/processengine/processengine.repository.ts
```diff
@@ -69,6 +69,22 @@ export class ProcessEngineRepository implements IProcessEngineRepository {
     return throwOnErrorResponse<any>(response);
   }
 
+  public async createProcessfromXML(xml: string): Promise<any> {
+    const xmlString: string = `{"xml": ${JSON.stringify(xml)}}`;
```
LeoTT: You could also use

```typescript
JSON.stringify({
    xml: xml,
});
```
easier to read, because it allows linebreaks and needs no template strings.

src/modules/processengine/processengine.repository.ts
```diff
@@ -69,6 +69,22 @@ export class ProcessEngineRepository implements IProcessEngineRepository {
...
+      headers: new Headers({
+        'Content-Type': 'application/json',
+      }),
+      body: xmlString,
+    };
+
+    const url: string = environment.processengine.routes.importBPMN;
+    const response: Response = await this.http.fetch(url, options);
+
+    return throwOnErrorResponse<any>(response);
```
LeoTT: Can we not use any for this case?
I would have no idea whether there is something useful inside this promise. I assume there is not, which would make it easy to get rid of this `any` and a couple of other `any`s which happened due to this.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -16,16 +17,32 @@ export class ProcessDefList {
   private getProcessesIntervalId: number;
   private subscriptions: Array<Subscription>;
 
+  @bindable()
+  private selectedFiles: FileList;
```
LeoTT: Should be public because it is accessed in the template-file. I know that typescript does not recognize aurelia-template files and therefore no errors will occur, but it would be a plus to be able to trust private vars being only used in that specific class.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -16,16 +17,32 @@ export class ProcessDefList {
...
+  private fileInput: HTMLInputElement;
```
LeoTT: same as above, public

electron_app/electron.js
```diff
@@ -40,8 +40,9 @@ getPort({port: 8000, host: '0.0.0.0'})
       minHeight: 800,
       icon: path.join(__dirname, '../build/icon.png'),  // only for windows and linux
     });
-
+
```
LeoTT: Delete whitespaces.
Most text editors allow you to delete trailing whitespaces end of lines after saving.

Vyperus:
When I import a process from my desktop this happens: 
bildschirmfoto 2018-04-17 um 09 43 53
Also I get this error:
bildschirmfoto 2018-04-17 um 09 45 05

Vyperus:
After updating all dependencies on the PE it works with other diagrams but not with the one I originally tested it with. I will look into this.

LeoTT:
If you import a diagram and there is already a process with that same name, it will simply overwrite that process definition.
That's a usability problem, which could be fixed in a later pr or now ('do you really want to overwrite..').

But it does more harm.
The database will be inconsistent. There will be elements referencing the now overwritten process definition.
Not clear what the damage is but can't be good.
Gives you the error message `Unhandledpromiserejectionwarning: Unhandled promise rejection (rejection id: 2): TypeError: Cannot read property 'data' of null` because the error handling in the postgres adapter is not helpful. 
That problem should be fixed before merging, probably with an error message like `There already is a process definition with that name.`

Vyperus:
Here are some informations I was able to extract from the xml's:

- if the diagram has NO lane, the import will work
- if the diagram has a lane but the name is not set, the import will break
- if the name of the lane is set, the property-panel can't read the id of the processRef if you click on the participant
- the name of the participant is not displayed correctly

In general: Diagrams that were exported by BPMN-Studio seem to work, but the import can't handle anything that isn't expected by us. Please check this again!

Vyperus:
Me and @alexanderkasten solved the issues I mentioned above and are now working to prevent @LeoTT issues from happening

src/modules/processengine/processengine.repository.ts
```diff
@@ -88,10 +91,19 @@ export class ProcessEngineRepository implements IProcessEngineRepository {
...
+      const url: string = environment.processengine.routes.importBPMN;
+      const response: Response = await this.http.fetch(url, options);
+      result = await throwOnErrorResponse<IErrorResponse>(response);
+    } catch (error) {
+      importError = error;
+    }
 
-    return throwOnErrorResponse<IErrorResponse>(response);
+    if (importError) {
+      throw importError;
```
LeoTT: I don't get this, you caught the error, saved it to a local var, then threw the error again. 
Is there a benefit? I don't see it and you could get rid of about 10 lines of code in that function.

src/modules/processengine/processengine.repository.ts
```diff
@@ -69,6 +75,30 @@ export class ProcessEngineRepository implements IProcessEngineRepository {
...
+        xml: xml,
+      }),
+    };
+
+    try {
+      const url: string = environment.processengine.routes.importBPMN;
+      const response: Response = await this.http.fetch(url, options);
+      result = await throwOnErrorResponse<IErrorResponse>(response);
+    } catch (error) {
+      throw error;
```
LeoTT: Looks better, but not quite what I meant. Why use try/catch? If something causes an error you gain nothing from treating that error by catching it and throwing it again.

If you don't want to do anything in that catch block you can just leave the error untreated and delete try/catch and make result a const.

PR #263, opened 2018-04-23
✨ Add Feature/use notificationservice instead toastr

### LINKED ISSUE #255: Umziehen der Toastr Notifications auf den Notification Service
#### Describe your issue

Es wurde ein NotificationService hinzugefügt, der in der `app.ts` Toastr übergeben bekommt.
Überall wo Toastr benutzt wird sollte nun auf den NotificationService gewechselt werden.

Der NotificationService soll auch erweitert werden mit:
- Types, Enums(für verschiedene Notifications)

#### Issue checklist

- [ ] I've included all the information that i think is relevant

alexanderkasten:
## What did you change?

This PR uses the notification service instead of toastr.
Closes #255 
Before merging the file association PR #241 should get merged.

- checkout the branch
- `npm i``
- e.g navigate to the detail view of a process and save the diagram

## PR-Checklist


src/app.ts
```diff
@@ -1,11 +1,20 @@
...
+@inject('NotificationService')
 export class App {
 
   public router: Router;
   public environment: any = environment;
+  public notificationService: NotificationService;
+
+  constructor(notificationService: NotificationService) {
+    this.notificationService = notificationService;
+    this.notificationService.setToastrInstance(toastr);
```
LeoTT: Why is this necessary? Initializing toastrInstance from here instead of importing in notification-service.ts?

LeoTT:
Why is `notificationService` a class?
Usage would be more intuitive and have less overhead if it was a function.
```javascript
import {NotificationType, notify} from '.....';

....
notify(NotificationType.ERROR, 'this is an error');
....
```
Right now you need to inject the notificationservice with the magic string 'NotificationService', declare a private member for the service, initialize the private member and use it to access the function in every class which wants to just call a function.

alexanderkasten:
@5ebastianMeier what do you think about Leo's comment ?

src/modules/notification/notification.service.ts
```diff
@@ -0,0 +1,57 @@
+import {INotification, NotificationType} from '../../contracts/index';
```
LeoTT:
```ts
import {Toastr} from 'toastr';
```
type import missing

alexanderkasten:
@LeoTT The Toastr instance is set in the app.ts, because if a module is trying to show up a notification before the toastr instance is being set, the message would be saved in an array and would showed up after the instance is set up.
If the toastr instance is set in the notification.service.ts, the function to show the notification would only be functional when the class is registered with aurelia.

LeoTT:
I don't get why initializing that in the constructor of `app.ts` would make a difference.
Where is the benefit?

Are you sure this is about toastr being initialized? Just tested it and I could use it right away (and I would find a notificationservice which requires startup time odd).
Funny behaviour is that aurelias startup seems to "swallow" messages, which are called before the dom is loaded. Is that what this is about?

If so there is no need to import 'toastr' and set it in `app.ts`.
You could just use a boolean `domIsLoaded` and import toastr in notification service right away.

But that's still some weird relation between `app` and `notification.service`.

How about this in notification.service instead?
```js
...
  constructor(ea: EventAggregator) {
    ea.subscribe('router:navigation:complete', () => {
   ... display all buffered messages here ...
    });
...
```

Would still find it neat if the result was not a class but a function.

alexanderkasten:
@LeoTT 

> Funny behaviour is that aurelias startup seems to "swallow" messages, which are called before the dom is loaded. Is that what this is about?

Yes it is. The notificationService gets the toastr instance via the `setToastrInstance` method in app.ts when the DOM is loaded. So the messages, that are "swallowed" from aurelia while starting up, will now be showed after set the toastr instance.

I have tested your example with importing toastr in notification.service.ts and subscribing on `router:navigation:complete` with the EventAggregator and it works. Now I will remove it from app.ts.

LeoTT:
@alexanderkasten sounds good.
Make sure to dispose the subscription afterwards, because that event will not only be fired once (and you only need the first time).

PR #272, opened 2018-04-27
✨ 🐛 Add Feature/fix condition on flow

### LINKED ISSUE #269: Default Flow allows condition
#### Describe your issue
At the current state, you can change the property of a flow from a normal to a so called `default flow`.

Per definition, a default Flow (or Path) is always taken, when there is no other path, that a token could take. 

An example might be a xor split, where no condition matches the actual token value. In this case, the token would take the path of the default flow. You can compare this with some sort of `default` in a `switch ... case` statement. 

In the current implementation, you can assign a condition to the default flow, which breaks the bpmn specification. 

screen shot 2018-04-26 at 17 12 35

* Select the flow on click on the wrench
* In the menu, click on "default Flow"
screen shot 2018-04-26 at 17 07 34

#### Possible solution
Remove the option to set a default flow. 

#### Issue checklist

alexanderkasten:
## What did you change?

This PR handles default-flows correctly. Closes #269 
Also it fixes a small bug with the condition value.

- checkout the branch
- `au run`
- click on normal flows, and then default-flows and see that you have no condition on the property panel

## PR-Checklist

### DIFF
--- a/src/contracts/bpmnmodeler/bpmnElements/IFlowElement.ts
+++ b/src/contracts/bpmnmodeler/bpmnElements/IFlowElement.ts
@@ -3,4 +3,5 @@ import {IModdleElement} from './IModdleElement';
 export interface IFlowElement extends IModdleElement {
   targetRef?: IModdleElement;
   sourceRef?: IModdleElement;
+  default?: IModdleElement;
 }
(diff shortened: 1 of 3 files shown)


robpal94:
Seems to work for me 👍 .

PR #276, opened 2018-05-04
✨ Add Feature/add timer event

### LINKED ISSUE #248: Implement Support for Timerevents in the Property Panel
#### Describe your issue

Timerevents are not supported by the Property Panel.

#### Possible solution

Implement support for the Timerevents. 
Orientate on the camunda property-panel.

alexanderkasten:
## What did you change?

This PR adds TimerEvent support to the PropertyPanel

Closes #248 

- checkout the branch
- `au run`
- add a timer boundary event to your diagram and set date, duration or cycle with the property panel

## PR-Checklist

### DIFF
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
@@ -1,0 +1,34 @@
+<template>
+  <require from="./timer-event.css"></require>
+  <require from="../../../../styles/sections.css"></require>
+  <div class="panel panel--grow">
+    <div class="panel__heading">
+      Timer Event
+    </div>
+    <div class="panel__content">
+      <table class="props-table">
+        <tr>
+          <td>
+            <select class="props-input" value.bind="selectedId" change.delegate="updateTimerType()">
+              <option model.bind="undefined">-Choose Option-</option>
+              <option model.bind="1">Date</option>
+              <option model.bind="2">Duration</option>
+              <option model.bind="3">Cycle</option>
+            </select>
+          </td>
+        </tr>
+      </table>
+      <table show.bind="selectedId" class="props-table probs-table--details">
+        <tr>
+          <th>Timer Definition</th>
+          <td>
+            <input type="text" class="props-input" value.bind="timerElement.body" change.delegate="updateTimerDefinition()">
+          </td>
+          <td>
+            <span class="glyphicon glyphicon-remove-circle props-table--clear-button" click.delegate="clearDefinition()"></span>
+          </td>
+        </tr>
+      </table>
+    </div>
+  </div>
+</template>
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
@@ -18,7 +18,7 @@
           </td>
         </tr>
       </table>
-      <table show.bind="timerType" class="props-table probs-table--details">
+      <table show.bind="timerType !== undefined" class="props-table probs-table--details">
         <tr>
           <th>Timer Definition</th>
           <td>
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
@@ -11,9 +11,9 @@
           <td>
             <select class="props-input" value.bind="timerType" change.delegate="updateTimerType()">
               <option model.bind="undefined">-Choose Option-</option>
-              <option model.bind="1">Date</option>
-              <option model.bind="2">Duration</option>
-              <option model.bind="3">Cycle</option>
+              <option model.bind="TimerType.Date">Date</option>
+              <option model.bind="TimerType.Duration">Duration</option>
+              <option model.bind="TimerType.Cycle">Cycle</option>
             </select>
           </td>
         </tr>
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
@@ -9,7 +9,7 @@
       <table class="props-table">
         <tr>
           <td>
-            <select class="props-input" value.bind="selectedId" change.delegate="updateTimerType()">
+            <select class="props-input" value.bind="timerType" change.delegate="updateTimerType()">
               <option model.bind="undefined">-Choose Option-</option>
               <option model.bind="1">Date</option>
               <option model.bind="2">Duration</option>
@@ -18,7 +18,7 @@
           </td>
         </tr>
       </table>
-      <table show.bind="selectedId" class="props-table probs-table--details">
+      <table show.bind="timerType" class="props-table probs-table--details">
         <tr>
           <th>Timer Definition</th>
           <td>
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
@@ -94,6 +94,7 @@ export class TimerEventSection implements ISection {
     };
 
     Object.assign(this.businessObjInPanel.eventDefinitions[0], timerTypeObject);
+    this.timerElement.body = '';
   }
 
   public updateTimerDefinition(): void {
--- a/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
+++ b/src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
@@ -1,0 +1,107 @@
+import {
+  IBpmnModdle,
+  IBpmnModeler,
+  IDefinition,
+  IElementRegistry,
+  IEscalation,
+  IEscalationElement,
+  IModdleElement,
+  IPageModel,
+  ISection,
+  IShape,
+} from '../../../../../../contracts';
+import {ITimerElement} from './../../../../../../contracts/bpmnmodeler/bpmnElements/ITimerElement';
+
+export class TimerEventSection implements ISection {
+
+  public path: string = '/sections/timer-event/timer-event';
+  public canHandleElement: boolean = false;
+
+  private businessObjInPanel: IModdleElement;
+  private moddle: IBpmnModdle;
+  private modeler: IBpmnModeler;
+
+  private escalationCodeVariable: string;
+  private isBoundaryEvent: boolean = true;
+
+  public selectedId: string;
+  public timerElement: any;
+
+  public async activate(model: IPageModel): Promise<void> {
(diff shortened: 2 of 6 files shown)


src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+  public path: string = '/sections/timer-event/timer-event';
+  public canHandleElement: boolean = false;
+
+  private businessObjInPanel: IModdleElement;
+  private moddle: IBpmnModdle;
+  private modeler: IBpmnModeler;
+
+  private isBoundaryEvent: boolean = true;
+
+  public selectedId: number;
```
LeoTT: Please use an enum instead of ids and a different variable name.

```ts
enum TimerType {
  Date = 'date',
  Duration = 'duration',
  Cycle = 'cycle',
}

...
// next line is required to expose the enum to aurelia view.
  private selectedTimerType: typeof TimerType = TimerType;
...
```

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+  private _init(): void {
+    const elementHasNoTimeDate: boolean = this.businessObjInPanel.eventDefinitions[0].timeDate === undefined;
+    const elementHasNoTimeDuration: boolean = this.businessObjInPanel.eventDefinitions[0].timeDuration === undefined;
+    const elementHasNoTimeCycle: boolean = this.businessObjInPanel.eventDefinitions[0].timeCycle === undefined;
+
+    if (elementHasNoTimeCycle && elementHasNoTimeDate && elementHasNoTimeDuration) {
+      return;
+    }
+
+    if (!elementHasNoTimeCycle) {
```
LeoTT: I'm a slow thinker which is why I like to avoid double negation.
Might be a matter of taste, if you don't like you can disregard this but in this case I would leave out the boolean vars.
```ts
const {timeDate, timeDuration, timeCycle} = this.businessObjInPanel.eventDefinitions[0];
if ((timeDate === undefined) 
  && (timeDuration === undefined) 
  && (timeCycle === undefined)) {
  return;
}
if (timeCycle !== undefined) {
```

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+      this.selectedId = 3; // tslint:disable-line:no-magic-numbers
+    } else if (!elementHasNoTimeDuration) {
+      this.selectedId = 2; // tslint:disable-line:no-magic-numbers
+    } else if (!elementHasNoTimeDate) {
+      this.selectedId = 1;
+    } else {
```
LeoTT: This last case never happens (because of that early return condition before).

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (this.businessObjInPanel.eventDefinitions[0].timeDuration !== undefined) {
+      return this.businessObjInPanel.eventDefinitions[0].timeDuration;
+    } else if (this.businessObjInPanel.eventDefinitions[0].timeDate !== undefined) {
+      return this.businessObjInPanel.eventDefinitions[0].timeDate;
+    } else if (this.businessObjInPanel.eventDefinitions[0].timeCycle !== undefined) {
+      return this.businessObjInPanel.eventDefinitions[0].timeCycle;
+    } else {
+      const timerEventDefinition: IModdleElement = this.moddle.create('bpmn:FormalExpression', {body: ''});
+      return timerEventDefinition;
+    }
```
LeoTT: Looks nicer with object destructuring and leaving out else (which is not needed because of early returns).

```ts
const {timeDuration, timeDuration, timeDate} = this.businessObjInPanel.eventDefinitions[0];

if (timeDuration !== undefined) {
   return timeDuration;
}
if (timeDate !== undefined) {
  return timeDate;
}
if (timeCycle !== undefined) {
  return timeCycle;
}

const timerEventDefinition: IModdleElement = this.moddle.create('bpmn:FormalExpression', {body: ''});
return timerEventDefinition;
```

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public isSuitableForElement(element: IShape): boolean {
+    if (this._elementIsTimerEvent(element)) {
+      this.isBoundaryEvent = this._elementIsBoundaryEvent(element);
```
LeoTT: Don't know what's happening here but from a public function with a name like `isSuitableForElement` I would not expect a side effect.

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+    const elementHasNoTimeDuration: boolean = this.businessObjInPanel.eventDefinitions[0].timeDuration === undefined;
+    const elementHasNoTimeCycle: boolean = this.businessObjInPanel.eventDefinitions[0].timeCycle === undefined;
+
+    if (elementHasNoTimeCycle && elementHasNoTimeDate && elementHasNoTimeDuration) {
+      return;
+    }
+
+    if (!elementHasNoTimeCycle) {
+      this.selectedId = 3; // tslint:disable-line:no-magic-numbers
+    } else if (!elementHasNoTimeDuration) {
```
LeoTT: You could use an early return statement instead of `else if`.

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private _elementIsBoundaryEvent(element: IShape): boolean {
+    return element !== undefined
+        && element.businessObject !== undefined
+        && element.businessObject.$type === 'bpmn:BoundaryEvent';
+  }
+
+  public updateTimerType(): void {
+    switch (this.selectedId) {
```
LeoTT: Got 2 ways to make this function nicer. Used `this.selectedId` in both cases as it currently is, would prefer it to be an enum (as mentioned in another comment).
```ts
const {timeDuration, timeCycle, timeDate} = this.businessObjInPanel.eventDefinitions[0];
const moddleElement = this.moddle.create('bpmn:FormalExpression', {body: this.timerElement.body});
timeDuration = undefined;
timeCycle = undefined;
timeDate = undefined;
switch (this.selectedId) {
  case 1:
    timeDate = moddleElement;
    break;
  case 2:
    timeDuration = moddleElement;
    break;
  case 3:
   timeCycle = moddleElement;
  default:
   break;
}
```

or even shorter(reduces 30+ loc to non cryptic 5 loc):
```ts
const moddleElement = this.moddle.create('bpmn:FormalExpression', {body: this.timerElement.body});
this.businessObjInPanel.eventDefinitions[0] = {
  timeDate: (this.selectedId === 1)  ? moddleElement : undefined,
  timeDuration: (this.selectedId === 2)  ? moddleElement : undefined,
  timeCycle: (this.selectedId === 3) ? moddleElement : undefined,
}
```
Functionality is a tiny bit different because in the lower code it replaces the whole object instead of the properties. Does the same unless other classes expect a reference to`this.businessObjInPanel.eventDefinitions[0]` to always be up to date.

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
```diff
@@ -0,0 +1,34 @@
...
+    <div class="panel__heading">
+      Timer Event
+    </div>
+    <div class="panel__content">
+      <table class="props-table">
+        <tr>
+          <td>
+            <select class="props-input" value.bind="timerType" change.delegate="updateTimerType()">
+              <option model.bind="undefined">-Choose Option-</option>
+              <option model.bind="1">Date</option>
```
LeoTT: You could do this:
```html
<option model.bind="TimerType.Date">Date</option> 
```
if you use the following instance var in `timer-event.ts`:
```ts
  public TimerType: typeof TimerType = TimerType;
```

Not a huge fan of having to add a class member for this but it's handy having access to enums in html. The whole point of enums is to get rid of magic strings/numbers and the mixture of enum values and number values for the same variable is worse than no enums at all.

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,113 @@
...
+  IModdleElement,
+  IPageModel,
+  ISection,
+  IShape,
+} from '../../../../../../contracts';
+
+enum TimerType {
+  Date = 1,
+  Duration = 2,
+  Cycle = 3,
```
LeoTT: TypeScript does this automatically (at least almost, starts with 0 instead of 1).
If you expose the enum to the view(see other comment) you could just use:
```ts
enum TimerType {
  Date,
  Duration,
  Cycle,
}
```
If you want the meaning of the values to be recognizable during runtime (for debugging for example) you could alternatively use string enums instead.

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (timeCycle !== undefined) {
+      return timeCycle;
+    }
+
+    const timerEventDefinition: IModdleElement = this.moddle.create('bpmn:FormalExpression', {body: ''});
+    return timerEventDefinition;
+  }
+
+  public isSuitableForElement(element: IShape): boolean {
+    if (this._elementIsTimerEvent(element)) {
```
LeoTT:
```ts
return this._elementIsTimerEvent(element);
```

src/modules/property-panel/indextabs/general/sections/timer-event/timer-event.html
```diff
@@ -0,0 +1,34 @@
...
+              <option model.bind="TimerType.Cycle">Cycle</option>
+            </select>
+          </td>
+        </tr>
+      </table>
+      <table show.bind="timerType" class="props-table probs-table--details">
+        <tr>
+          <th>Timer Definition</th>
+          <td>
+            <input type="text" class="props-input" value.bind="timerElement.body" change.delegate="updateTimerDefinition()">
```
LeoTT: This can only handle specific input (depending on timerType).
Here's are the [formats](https://docs.camunda.org/manual/7.6/reference/bpmn20/events/timer-events/).
Not sure if there is a reason to not check for the format(camunda doesn't do it). Does not have to be in this pr, but might be worth an issue.

PR #278, opened 2018-05-07
✨Add Feature/add conditional event

alexanderkasten:
## What did you change?

Now the property panel can handle conditional boundary events (used in Reservierungsprozess.bpmn). 

- attach a boundary event to a task (e.g. servicetask) 
- change the boundary event type to conditional boundary event
- look at the property panel

## PR-Checklist

### DIFF
--- a/src/modules/property-panel/indextabs/general/general.ts
+++ b/src/modules/property-panel/indextabs/general/general.ts
@@ -1,6 +1,7 @@
 import {IIndextab, IPageModel, ISection, IShape} from '../../../../contracts';
 import {BasicsSection} from './sections/basics/basics';
 import {CallActivitySection} from './sections/call-activity/call-activity';
+import {ConditionalEventSection} from './sections/conditional-event/conditional-event';
 import {ErrorEventSection} from './sections/error-event/error-event';
 import {EscalationEventSection} from './sections/escalation-event/escalation-event';
 import {FlowSection} from './sections/flow/flow';
@@ -23,6 +24,7 @@ export class General implements IIndextab {
   public flowSection: ISection = new FlowSection();
   public errorEventSection: ISection = new ErrorEventSection();
   public escalationEventSection: ISection = new EscalationEventSection();
+  public conditionalEventSection: ISection = new ConditionalEventSection();
 
   public sections: Array<ISection> = [
     this.basicsSection,
@@ -34,6 +36,7 @@ export class General implements IIndextab {
     this.flowSection,
     this.errorEventSection,
     this.escalationEventSection,
+    this.conditionalEventSection,
   ];
 
   public canHandleElement: boolean = true;
--- a/src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
+++ b/src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
@@ -31,27 +31,14 @@ export class ConditionalEventSection implements ISection {
 
     const {variableName, variableEvent} = this.businessObjInPanel.eventDefinitions[0];
 
-    if (variableEvent !== undefined) {
-      this.variableEvent = this.businessObjInPanel.eventDefinitions[0].variableEvent;
-    } else {
-      this.variableEvent = '';
-    }
-
-    if (variableName !== undefined) {
-      this.variableName = this.businessObjInPanel.eventDefinitions[0].variableName;
-    } else {
-      this.variableName = '';
-    }
-
+    this.variableEvent = (variableEvent === undefined) ? '' : variableEvent;
+    this.variableName = (variableName === undefined) ? '' : variableName;
   }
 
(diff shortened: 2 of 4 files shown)


src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
```diff
@@ -0,0 +1,90 @@
...
+  private _init(): void {
+    this.conditionalElement = this._getConditionalElement();
+
+    const {variableName, variableEvent} = this.businessObjInPanel.eventDefinitions[0];
+
+    if (variableEvent !== undefined) {
+      this.variableEvent = this.businessObjInPanel.eventDefinitions[0].variableEvent;
+    } else {
+      this.variableEvent = '';
+    }
```
LeoTT:
```ts
this.variableEvent = (variableEvent === undefined) ? '' : variableEvent;
```

src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    if (variableName !== undefined) {
+      this.variableName = this.businessObjInPanel.eventDefinitions[0].variableName;
+    } else {
+      this.variableName = '';
+    }
```
LeoTT: see comment above

src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  }
+
+  private _getConditionalElement(): IModdleElement {
+    if (this.businessObjInPanel.eventDefinitions[0].condition !== undefined) {
+      return this.businessObjInPanel.eventDefinitions[0].condition;
+    } else {
```
LeoTT: no else needed due to return.

```ts
const {condition} = this.businessObjInPanel.eventDefinitions[0];
const conditionalEventDefinition: IModdleElement = this.moddle.create('bpmn:FormalExpression', {body: ''});
return (condition === undefined) ? conditionalEventDefinition : condition;
```

src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const {variableName, variableEvent} = this.businessObjInPanel.eventDefinitions[0];
+
+    this.variableEvent = (variableEvent === undefined) ? '' : variableEvent;
+    this.variableName = (variableName === undefined) ? '' : variableName;
+  }
+
+  private _getConditionalElement(): IModdleElement {
+    const {condition} = this.businessObjInPanel.eventDefinitions[0];
+    const conditionalEventDefinition: IModdleElement = this.moddle.create('bpmn:FormalExpression', {body: ''});
```
LeoTT: What happens to this new moddleElement? 
For example in the following function:
```ts
  public updateConditionalElement(): void {
    const conditionalElement: IModdleElement = this._getConditionalElement();
    conditionalElement.body = this.conditionalElement.body;
  }
```
The private  _get function creates the new element, you add something to its property `body` and then nothing. New object with updated `body` is never to be seen again. 
Is something like `this.businessObjInPanel.eventDefinitions[0].condition = conditionalEventDefinition;` missing?

Another thing, I expect a function starting with `_get` to just get me something and not do any sideeffects. Would be better imo if you created the new Moddleelement in the update function or better yet (if it makes sense) in the init function.

src/modules/property-panel/indextabs/general/sections/conditional-event/conditional-event.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class ConditionalEventSection implements ISection {
+
+  public path: string = '/sections/conditional-event/conditional-event';
+  public canHandleElement: boolean = false;
+
+  private businessObjInPanel: IModdleElement;
+  private moddle: IBpmnModdle;
+
+  public conditionalElement: IModdleElement;
```
LeoTT: Why bind with an object instead of a string? You don't want to bind to the original IModdleElement directly anyway.
In a function `conditionUpdated` you could use the string for the update just as well.

src/modules/property-panel/indextabs/general/general.ts
```diff
@@ -24,6 +25,7 @@ export class General implements IIndextab {
   public flowSection: ISection = new FlowSection();
   public errorEventSection: ISection = new ErrorEventSection();
   public escalationEventSection: ISection = new EscalationEventSection();
+  public conditionalEventSection: ISection = new ConditionalEventSection();
```
LeoTT: @Vyperus what's the idea behind this pattern?
Why not use normal aurelia components?

cmg-dev:
#292 includes the cleaned up features.

PR #294, opened 2018-05-08
✨ Add Select Name On Import

SteffenKn:
## What did you change?

A user can now choose a name for the process definition he is importing.

Instead of simply overwriting process definitions with the same name as the process definition to be imported, the user can now set a different name or intentionally overwrite the existing process definition.

- Add Select Own Name On Import
- Add Name Selection Modal
- Add Overwrite Confirmation Modal
- Add Name Suggestion

#### Code:

- Format Code of ProcessDef-List
- Fix Private Names
- Sort Imports and Members
- Refactor ProcessSolutionPanel

#### Style:

- Cleanup & Beautify CSS
- Improve CSS Classes for ProcessDef-List Buttons
- Prettify CSS and HTML of ProcessSolutionExplorer
- Refactor CSS of ProcessDef-List

- Style HTML of ProcessDef-List

- Replace Glyphicons With Font Awesome Icons
- Use Correct Icon for Importing and Exporting

- Add Folds for Readability
- Add Missing Comma and Spaces
- Add TODOs for Refactoring
- Add Types

- Improve Naming

#### Fixes:

- Fix Importing Same File Multiple Times
- Fix Lint Errors
- Fix Typos

## Related to:

- https://github.com/process-engine/process_engine/pull/81
- https://github.com/process-engine/process_engine_http/pull/11
- https://github.com/process-engine/process_engine_contracts/pull/27

- Import a process definition
- Take a look at the modal with a text box where you can enter a name for the process definition.
- Import another process definition and enter the same name as before.
- Take a look at the modal, where you can choose between 3 options:
	1. overwrite the existing process definition
	2. change the name of the new process definition
	3. cancel the import

## PR-Checklist

- [ ] I've **not added any other changes** than the ones described above.

### DIFF
--- a/src/modules/processdef-list/processdef-list.html
+++ b/src/modules/processdef-list/processdef-list.html
@@ -12,7 +12,7 @@
             <div class="modal-header">
               <h1 class="modal-dialog__headline">Import BPMN Diagram</h1>
             </div>
-            <div class="modal-body">
+            <div class="modal-body process-def-list__modal-body">
               <form submit.delegate = "checkDiagrammName()">
                 Import as: <input type="text" value.bind="diagrammToImport.name" focus.one-time="true" placeholder="Definition Name">
               </form>
@@ -33,7 +33,7 @@
             <div class="modal-header">
               <h1 class="modal-dialog__headline">Diagramm already exists!</h1>
             </div>
-            <div class="modal-body">
+            <div class="modal-body process-def-list__modal-body">
               A diagramm with that name is already imported. Hit Cancel to change the name.
             </div>
             <div class="modal-footer">
--- a/src/modules/processdef-list/processdef-list.html
+++ b/src/modules/processdef-list/processdef-list.html
@@ -5,6 +5,7 @@
 
   <div class="process-def-list">
     <!-- Modal dialog for importing BPMN diagrams {{{ -->
+    <!-- Import Dialog {{{ -->
     <div if.bind="showDiagramNameDialog">
       <div class="modal fade in show" tabindex="-1" role="dialog">
         <div class="modal-dialog" role="document">
@@ -24,8 +25,12 @@
           </div>
         </div>
       </div>
+      <!-- Modal Gray Bacground {{{ -->
       <div class="modal-backdrop fade in"></div>
+      <!-- }}} Modal Gray Bacground -->
     </div>
+    <!-- }}} Import Dialog -->
+    <!-- Overwrite Dialog {{{ -->
     <div if.bind="showOverwriteDialog">
       <div class="modal  fade in show" tabindex="-1" role="dialog">
         <div class="modal-dialog" role="document">
@@ -45,8 +50,11 @@
           </div>
         </div>
       </div>
+      <!-- Modal Gray Bacground {{{ -->
       <div class="modal-backdrop fade in"></div>
+      <!-- }}} Modal Gray Bacground -->
     </div>
+    <!-- }}} Overwrite Dialog -->
     <!-- }}} Modal dialog for importing BPMN diagrams -->
 
     <!-- TODO: Refactor this...  this has to land in the router! {{{ -->
--- a/src/modules/processdef-list/processdef-list.html
+++ b/src/modules/processdef-list/processdef-list.html
@@ -102,10 +102,10 @@
     <!-- TODO: Refactor this - put into a component {{{ -->
     <div class="processdef-list-container">
       <div show.bind="_processes">
-        <a class="btn btn-success create-process" click.delegate="createProcess()">
+        <a class="btn btn-success processdef-list__btn-create" click.delegate="createProcess()">
           <i class="fa fa-magic"></i> Create Process Definition
         </a>
-        <a class="btn btn-primary import-bpmn" click.delegate="fileInput.click()">
+        <a class="btn btn-primary processdef-list__btn-import" click.delegate="fileInput.click()">
           <i class="fa fa-download"></i> Import BPMN
         </a>
         <input type="file" ref="fileInput" accept=".xml,.bpmn" files.bind="selectedFiles" class="hidden">
--- a/src/modules/processdef-list/processdef-list.html
+++ b/src/modules/processdef-list/processdef-list.html
@@ -11,8 +11,9 @@
             <div class="modal-header">
               <h5 class="modal-title">Import BPMN Diagram</h5>
             </div>
-            <form submit.delegate = "checkDiagrammName()">
-              <input type="text" class="import-process-textbox" value.bind="diagrammToImport.name" focus.one-time="true" placeholder="Definition Name">
+            <form class="modal-dialog__import-bpmn" submit.delegate = "checkDiagrammName()">
+              <p>Import as:</p>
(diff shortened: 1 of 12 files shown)


src/modules/processdef-list/processdef-list.ts
```diff
@@ -23,6 +23,8 @@ export class ProcessDefList {
   public selectedFiles: FileList;
   public fileInput: HTMLInputElement;
   private reader: FileReader = new FileReader();
+  public diagrammToOverride: any;
+  public diagrammToImport: any;
```
LeoTT: Avoid type `any`. It's better when you use no type at all instead, because:
- type inference still has a chance to do its magic if you don't use `any` (not here but in general)
- when I see `any` I expect that it can literally be anything (e.g. in a function for a logger which can log anything)

In this case the type seems to be:
```ts
{
   name: string,
   xml: string,
}
```

Based on code only I have really no idea what is happening with the two variables above.
Why do they even contain xml(when xml is managed within this class alone)?
Can `diagrammToImport` be instead just a string variable for the name of the new diagram?
Why is `diagrammToOverride` exposed to view? For the `if.bind`? boolean would be better there.
What is `diagrammToOverride` good for? You already got the information(name, xml), right?

src/modules/processdef-list/processdef-list.html
```diff
@@ -4,6 +4,46 @@
...
 
   <div class="process-def-list">
+    <div if.bind="diagrammToImport">
+      <div class="modal  fade in show" tabindex="-1" role="dialog">
+        <div class="modal-dialog" role="document">
+          <div class="modal-content">
+            <div class="modal-header">
+              <h5 class="modal-title">Import BPMN Diagram</h5>
+            </div>
+            <form class="modal-dialog__import-bpmn" submit.delegate = "checkDiagrammName()">
```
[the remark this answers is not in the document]
SteffenKn: This gets called when you hit the return key in the textbox.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,58 +1,139 @@
...
+      return;
+    }
+
+    const isNameUnique: boolean = await this.checkIfProcessDefNameUnique(diagram.name);
+    if (!isNameUnique) {
+      this.showOverwriteDialog = true;
+      return;
+    }
+
+    this._importProcess(diagram.name, diagram.xml);
```
LeoTT: I would expect a function starting with `check` to only do the checking and not administer the import.
Maybe turn this around, have a function called  `importDiagram` use a private check-function?

src/modules/processdef-list/processdef-list.html
```diff
@@ -43,24 +116,32 @@
...
+          <i class="fa fa-download"></i> Import BPMN
         </a>
         <input type="file" ref="fileInput" accept=".xml,.bpmn" files.bind="selectedFiles" class="hidden">
         <table class="table table-striped table-bordered">
           <tr>
-            <th>ProcessDefinitionName</th>
-            <th>ProcessDefinitionId</th>
+            <th>Process Defintion Name</th>
+            <!-- TODO: Drop the Id; it does not seem usefull -->
+            <th>Process Defintion Id</th>
```
LeoTT: typos in the two lines above: Defintion

src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,58 +1,150 @@
...
   public selectedFiles: FileList;
   public fileInput: HTMLInputElement;
-  private reader: FileReader = new FileReader();
+  public showOverwriteDialog: boolean;
+  public showDiagramNameDialog: boolean;
+  // TODO: Put this into an interface IBpmnDiagramm and into the contracts folder
+  public newDiagram: {
+                        name: string,
+                        xml: string,
+                      };
```
LeoTT: The only thing the view is accessing is the name. Imported xml could be a saved in a private variable.
Current solution has the disadvantage that it looks like xml is used with value-binding from here and that `value.bind=newDiagram.name` only works as long as you don't replace the object `newDiagram`.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,58 +1,150 @@
...
+    const processes: IPagination<IProcessDefEntity> = await this._bpmnStudioClient.getProcessDefList();
+    const existingProcess: IProcessDefEntity = processes.data.find((process: IProcessDefEntity) => {
+      return process.name === processDefName;
+    });
+
+    if (existingProcess === undefined) {
+      return true;
+    }
+
+    return false;
```
LeoTT: `return (existingProcess === undefined)`

src/modules/processdef-list/processdef-list.html
```diff
@@ -4,6 +4,77 @@
...
+                <label for="processDefinitionName">Import as:</label>
+                <input type="text" class="form-control"
+                  id="processDefinitionName"
+                  aria-describedby="processDefinitionNameHelp"
+                  value.bind="newDiagram.name"
+                  focus.one-time="true"
+                  placeholder="Definition Name">
+
+                <small id="processDefinitionNameHelp" class="form-text text-muted">
+                  This text will identify the .bpmn file are about to import.
```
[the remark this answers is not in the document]
SteffenKn: Done

src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,58 +1,142 @@
...
+
+      if (!isDiagramNameUnique) {
+        this.showOverwriteDialog = true;
+        return;
+      }
+
+      this._saveNewDiagram();
+    } catch (error) {
+      this._notificationService.showNotification(NotificationType.ERROR, 'Name can not be empty');
+      return;
```
LeoTT: Make the try-block as small as possible. If by chance `_saveNewDiagram` for example throws an error it will give you a notification about the name, which is irritating.
```ts
let isDiagramNameUnique: boolean;
try {
  isDiagramNameUnique = await this._isDiagramNameUniqueOrEmpty();
} catch (error) {
  this._notificationService.showNotification(NotificationType.ERROR, 'Name can not be empty');
  return;
}

...rest here
```

Alternatively, if this is the only place where that error might occur it's a bit nicer to do just the following at the beginning of this function.
```ts
if (this.newDiagramName === '' || this.newDiagramName === undefined) {
  this._notificationService.showNotification(NotificationType.ERROR, 'Name can not be empty');
  return;
}
```
Sorry, I recommended the try/catch in a previous comment.

src/modules/processdef-list/processdef-list.ts
```diff
@@ -1,58 +1,142 @@
...
+    this.showDiagramNameDialog = true;
+  }
+
+  public async checkIfProcessDefNameUnique(processDefName: string): Promise<boolean> {
+    const processes: IPagination<IProcessDefEntity> = await this._bpmnStudioClient.getProcessDefList();
+    const existingProcess: IProcessDefEntity = processes.data.find((process: IProcessDefEntity) => {
+      return process.name === processDefName;
+    });
+
+    return (existingProcess === undefined);
```
LeoTT:
```ts
const processExists: boolean = processes.data.some((process: IProcessDefEntity) => {
      return process.name === processDefName;
});
return processExists;
```

PR #468, opened 2018-06-05
✨Remove elements with backspace

### LINKED ISSUE #352: Leinwand: auf Mac Elemente auch mit BACKSPACE entfernen
**Beschreibe das Problem:**

Im Moment muss der Nutzer auf seinem MacBook FN-BACKSPACE drücken, um ein Element zu entfernen (um die ENTF-Taste zu emulieren).

Auf dem Mac ist es üblich, dass man Elemente (auch) mit der BACKSPACE-Taste löschen kann.

alexanderkasten:
## What did you change?

This PR fixes #352 

- checkout the branch
- `npm run build`
- `npm run electron-start-dev`
- open detail-view of a process
- select one or more elements/shapes
- click backspace and see that they are deleted

## PR-Checklist

### DIFF
--- a/src/contracts/bpmnmodeler/IKeyboard.ts
+++ b/src/contracts/bpmnmodeler/IKeyboard.ts
@@ -1,8 +1,8 @@
 export interface IKeyboard {
   addListener(keyListenerFunction: Function): void;
-  hasModifier(modifiers: any): any;
-  isCmd(modifiers: any): any;
-  isShift(modifiers: any): any;
+  hasModifier(modifiers: KeyboardEvent): boolean;
+  isCmd(modifiers: KeyboardEvent): boolean;
+  isShift(modifiers: KeyboardEvent): boolean;
   bind(node: HTMLElement): void;
   getBinding(): HTMLElement;
   unbind(): void;
--- a/src/contracts/bpmnmodeler/IKeyboard.ts
+++ b/src/contracts/bpmnmodeler/IKeyboard.ts
@@ -1,0 +1,9 @@
+export interface IKeyboard {
+  addListener(keyListenerFunction: Function): void;
+  hasModifier(modifiers: any): any;
+  isCmd(modifiers: any): any;
+  isShift(modifiers: any): any;
+  bind(node: HTMLElement): void;
+  getBinding(): HTMLElement;
+  unbind(): void;
+}
--- a/src/modules/bpmn-io/bpmn-io.ts
+++ b/src/modules/bpmn-io/bpmn-io.ts
@@ -8,6 +8,8 @@ import {ElementDistributeOptions,
         IBpmnFunction,
         IBpmnModeler,
         IDefinition,
+        IEditorActions,
+        IKeyboard,
         IModdleElement,
         IModeling,
         IProcessDefEntity,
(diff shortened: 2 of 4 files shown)


src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -321,4 +325,19 @@ export class BpmnIo {
...
   }
+
+  private _addKeyboardListener(): void {
+    const keyboard: IKeyboard = this.modeler.get('keyboard');
+    const editorActions: IEditorActions = this.modeler.get('editorActions');
+    const backSpaceKeyCode: number = 8;
+
+    keyboard.addListener(removeSelectedElements);
+
+    function removeSelectedElements(key: number, modifiers: any): boolean {
```
LeoTT: Generally, use function statements instead of function declarations. The difference is that js hoists function declarations to the top of the function block.

Are you able to remove the listener in `detached`? 

Also, no `any` please.

src/contracts/bpmnmodeler/IKeyboard.ts
```diff
@@ -0,0 +1,9 @@
+export interface IKeyboard {
+  addListener(keyListenerFunction: Function): void;
+  hasModifier(modifiers: any): any;
+  isCmd(modifiers: any): any;
+  isShift(modifiers: any): any;
```
LeoTT: Don't use `any` here. I suppose `isShift` and `isCmd` return `boolean`.
If you don't know what type something is, I recommend to not use a type at all. Better than `any`.

alexanderkasten: https://github.com/process-engine/bpmn-studio/issues/469

robpal94: In my understanding, the *no any* rule only applies on constants but not on methods, that can return a primitive value. For example when I read a method called `isCmd()` I would expect, that this method returns only a `boolean` value and nothing else. Despite the *any rule*, I would suggest to add a return type on every method, that does return a primitive type. Correct me if I'm wrong.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -321,4 +325,18 @@ export class BpmnIo {
       this._eventAggregator.publish(environment.events.processDefDetail.saveDiagram);
     }
   }
+
+  private _addKeyboardListener(): void {
+    const keyboard: IKeyboard = this.modeler.get('keyboard');
+    const editorActions: IEditorActions = this.modeler.get('editorActions');
+    const backSpaceKeyCode: number = 8;
```
robpal94: The keycode property is deprecated. See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

I think thats because the key code is not consistent in different browsers.

robpal94:
Just a general question: How is the behaviour of the implemented changes on machines, that offers a dedicated `del` Key? I would not expect that I can remove elements using backspace when Im on a Windows machine.

src/contracts/bpmnmodeler/IKeyboard.ts
alexanderkasten: @roblen45 @Vyperus I do not know the return type of the methods, so for now it will be `any`.

PR #491, opened 2018-06-18
♻️Refactor the Plattform check

robpal94:
## What did you change?
This PR moves the check, if the current Plattform is a mac, to a dedicated method. 

Since three methods use the same method to ask for the current platform, I thought that its a good idea to move the duplicated lines of code to a dedicated method.

This may also become handy, if we want to implement more features, that uses the meta key since we don't have to copy the same regex for the platform check over and over. 

Also, this PR fixes a few typos in the word _platform_ and change the name of the `_addKeyboardListener()` Method, since the only purpose of this method was to bind an event to the backspace key, if the current platform is a mac.

## PR-Checklist

### DIFF
--- a/src/modules/bpmn-io/bpmn-io.ts
+++ b/src/modules/bpmn-io/bpmn-io.ts
@@ -422,9 +422,9 @@ export class BpmnIo {
   }
 
   /**
-   * Checks, if the current platform is a mac.
+   * Checks, if the current platform is a macOS.
    *
-   * @returns true, if the current plattform is a mac
+   * @returns true, if the current platform is macOS
    */
   private _checkIfCurrentPlatformIsMac = (): boolean => {
     const macRegex: RegExp = /.*mac*./i;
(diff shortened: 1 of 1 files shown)


src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -400,4 +398,17 @@ export class BpmnIo {
       this._eventAggregator.publish(environment.events.processDefDetail.printDiagram);
     }
   }
+
+  /**
+   * Checks, if the current platform is a mac.
+   *
+   * @returns true, if the current plattform is a mac
+   */
+  private _currentPlattformIsMac = (): boolean => {
```
ElRaptorus: I'd expect a name like `currentPlatformIsMac` to belong to a property, not to a function.

Better name to `checkIfCurrentPlatformIsMac`.

Also, its `platform` in english. 😉

cmg-dev:
bitte noch nicht mergen... das ist mir sehr unwichtig; kann in den nächsten release.

PR #493, opened 2018-06-18
✨ Persist and Load Last Property Panel Width in Local Storage

SteffenKn:
## What did you change?

When returning to the view/diagram, the width of the property panel is restored.

- Open the BPMN-Studio
- Go to the "Detail View"
- Resize the property panel
- Go to the "Plan View"
- Go back to the "Detail View"
- Notice that the property panel still has the resized size

## PR-Checklist

### DIFF
--- a/src/modules/bpmn-io/bpmn-io.ts
+++ b/src/modules/bpmn-io/bpmn-io.ts
@@ -156,11 +156,15 @@ export class BpmnIo {
     ];
 
     const previousPropertyPanelWidth: string = window.localStorage.getItem('propertyPanelWidth');
-    if (previousPropertyPanelWidth !== undefined) {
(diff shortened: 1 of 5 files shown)


src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -173,6 +178,8 @@ export class BpmnIo {
     for (const subscription of this._subscriptions) {
       subscription.dispose();
     }
+
+    window.localStorage.setItem('propertyPanelWidth', '' + this.propertyPanelWidth);
```
robpal94: Why we are parsing an `integer` from a string and convert the parsed `integer` back to `string` without doing anything with that parsed `integer` ? 🙂

PR #502, opened 2018-06-19
🐛 Fix saving of an invalid diagram via cmd + s

robpal94:
## What did you change?
This PR Fixes the bug, where you can save an invalid diagram while pressing `cmd + s` while the property pane was active. 

Fixes: #501 

* Create a diagram
* Create two elements with the same id
* See, that the saving is no longer possible

## PR-Checklist

### DIFF
--- a/src/modules/bpmn-io/bpmn-io.ts
+++ b/src/modules/bpmn-io/bpmn-io.ts
@@ -353,15 +353,17 @@ export class BpmnIo {
     * @see environment.events.processDefDetail.saveDiagram
     */
     const sKeyIsPressed: boolean = event.key === 's';
-    const userWantsToSave: boolean = metaKeyIsPressed && sKeyIsPressed;
+    const userDoesNotWantsToSave: boolean = !(metaKeyIsPressed && sKeyIsPressed);
 
-    if (userWantsToSave) {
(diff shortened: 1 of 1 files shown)


Paulomart:
IMO der is valid check should me moved to the place there the save event is handled.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -340,18 +347,21 @@ export class BpmnIo {
...
+    */
     const sKeyIsPressed: boolean = event.key === 's';
     const userWantsToSave: boolean = metaKeyIsPressed && sKeyIsPressed;
 
     if (userWantsToSave) {
       // Prevent the browser from handling the default action for CTRL + s.
       event.preventDefault();
-      this._eventAggregator.publish(environment.events.processDefDetail.saveDiagram);
+
+      if (this._diagramIsValid) {
```
alexanderkasten: you could early exit at the top of this method instead of this.

`if (!this._diagramIsValid) {
  return;
}`

robpal94: IMO an early out is not necessary here, because the following if statement excludes the _diagram is not valid_ case. Maybe @cmg-dev or @LeoTT can also state on this.

PR #517, opened 2018-07-03
Fix corrupt UserTask icon when exporting to an image or printing

robpal94:
## What did you change?
Fixes #515 

This PR replaces the usage of the canvg libary to render and exporting a svg to a canvas since it seems to be broken (as referenced [here](https://github.com/canvg/canvg/issues/577) ).

## How can others test the changes?
* Create a Process with a UserTask
* Export this Process as an image or print it

## PR-Checklist

Please check the boxes in this list after submitting your PR:

- [x] You can merge this PR **right now** (if not, please prefix the title with "WIP: ")
- [x] I've tested **all** changes included in this PR.
- [x] I've also reviewed this PR myself before submitting (e.g. for scrambled letters, typos, etc.)
- [x] I've merged the `develop` branch into my branch before finishing this PR.
- [x] I've **not added any other changes** than the ones described above.
- [x] I've mentioned all **PRs, which relate to this one**

PR #523, opened 2018-07-04
✨ Add DiagramExportService and PrintService

alexanderkasten:
## What did you change?

This PR adds a diagram export service and a print service.

- checkout the branch
- start bpmn-studio

## PR-Checklist


src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  public async exportSVG(process: IProcessDefEntity): Promise<void> {
+    const svg: string = await this.getSVG();
+
+    download(svg, `${process.name}.svg`, 'image/svg+xml');
+  }
+
+  public async exportPNG(process: IProcessDefEntity): Promise<void> {
+    const svg: string = await this.getSVG();
+    const imageURL: string = await this._generateImageFromSVG('png', svg);
```
robpal94: The `_drawSQLToCanvas` method, used by the `_generateImageFromSVG()` method rejects the promise, if the image could not be loaded. Since `_generateImageFromSVG()` does not handles this rejection, it will be passed to the `exportABC()` methods. 

This PR also adds a handling of those promise rejections, by displaying a little notification. https://github.com/process-engine/bpmn-studio/pull/517/files#diff-8639e153d8f9561c480d171796d13848R429

We should update this export methods to do this.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -120,9 +102,9 @@ export class ProcessDefDetail {
       this._eventAggregator.subscribe(environment.events.diagramChange, () => {
         this._diagramHasChanged = true;
       }),
-      this._eventAggregator.subscribe(environment.events.processDefDetail.printDiagram, () => {
-        this._printDiagram();
-      }),
+      // this._eventAggregator.subscribe(environment.events.processDefDetail.printDiagram, () => {
```
robpal94: I think we can remove this lines.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -236,13 +236,17 @@ export class BpmnIo {
 
   public async toggleXMLView(): Promise<void> {
     if (!this.showXMLView) {
-      this.xml = await this.getXML();
+      this.xml = await this._diagramExportService.getXML();
```
Vyperus: Why is this service used to get the xml from the diagram? The bpmn-io component should know about the xml anyway.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -236,13 +236,17 @@ export class BpmnIo {
...
       this.showXMLView = true;
     } else {
       this.showXMLView = false;
     }
   }
 
+  public getXML(): Promise<string> {
+    return this._diagramExportService.getXML();
```
Vyperus: Same here. The exportservice shouldn't handle the xml for the modeler that I have in this component anyway.

src/contracts/export/IDiagramExportService.ts
```diff
@@ -0,0 +1,10 @@
+import {IProcessDefEntity} from '../processengine';
+
+export interface IDiagramExportService {
+  exportBPMN(process: IProcessDefEntity): Promise<void>;
+  exportSVG(process: IProcessDefEntity): Promise<void>;
+  exportPNG(process: IProcessDefEntity): Promise<void>;
+  exportJPEG(process: IProcessDefEntity): Promise<void>;
+  getXML(): Promise<string>;
```
Vyperus: Why does the export service have a function called getXML()? If this service should handle the xml for us it would be called "XML Service" or something like this. Don't add extra functionality to service that doesn't need that functionality.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,189 @@
...
+   *
+   * @param svgWidth With of the diagrams canvas element.
+   * @param svgHeight Height of the diagrams canvas element.
+   * @param targetDPI DPI of the output.
+   * @param diagonalSize Diagonal Size of the printed document.
+   */
+  private _calculatePixelRatioForDPI(svgWidth: number, svgHeight: number, targetDPI: number, diagonalSize: number): number {
+
+    // tslint:disable:no-magic-numbers
+    const svgWidthSquared: number = Math.pow(svgWidth, 2);
```
Vyperus: So for what does the 2 stand for? 🤔

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,189 @@
+import * as download from 'downloadjs';
+import * as beautify from 'xml-beautifier';
+import {IBpmnModeler, IDiagramExportService, IProcessDefEntity} from '../../../contracts';
```
cmg-dev: Newline above...

Please sort the `imports` correctly.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -139,6 +148,21 @@ export class BpmnIo {
       this._eventAggregator.subscribe(environment.events.navBar.disableSaveButton, () => {
         this._diagramIsValid = false;
       }),
+      this._eventAggregator.subscribe(`${environment.events.processDefDetail.exportDiagramAs}:BPMN`, (process: IProcessDefEntity) => {
```
Vyperus: This won't work as expected since the button, which is firing the export events, won't know about the process that will be exported it just triggers the event.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -139,6 +148,21 @@ export class BpmnIo {
...
+        this._diagramExportService.exportBPMN(process);
+      }),
+      this._eventAggregator.subscribe(`${environment.events.processDefDetail.exportDiagramAs}:SVG`, (process: IProcessDefEntity) => {
+        this._diagramExportService.exportSVG(process);
+      }),
+      this._eventAggregator.subscribe(`${environment.events.processDefDetail.exportDiagramAs}:PNG`, (process: IProcessDefEntity) => {
+        this._diagramExportService.exportPNG(process);
+      }),
+      this._eventAggregator.subscribe(`${environment.events.processDefDetail.exportDiagramAs}:JPEG`, (process: IProcessDefEntity) => {
```
Vyperus: See above.

src/modules/navbar/navbar.ts
```diff
@@ -89,7 +89,7 @@ export class NavBar {
   }
 
   public exportDiagram(exportAs: string): void {
-    this._eventAggregator.publish(`${environment.events.processDefDetail.exportDiagramAs}:${exportAs}`);
+    this._eventAggregator.publish(`${environment.events.processDefDetail.exportDiagramAs}:${exportAs}`, this.process);
```
Vyperus: See my comment above. Right now the navbar knows what diagram is displayed but that could change in the future. Why not let the component that holds the diagram decide which to export?

src/modules/bpmn-io/services/diagram.export.service.ts
cmg-dev: @Vyperus  suggest something please ;)

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,189 @@
...
+import {NotificationType} from '../../../contracts/notification/constants';
+import {NotificationService} from '../../notification/notification.service';
+
+export class DiagramExportService implements IDiagramExportService {
```
cmg-dev: I would seriously like an interface like this:

```ts
diagramExportService
  .exportDiagram()
  .asXML();

diagramExportService
  .exportDiagram()
  .asSVG();

diagramExportService
  .exportDiagram()
  .asBPMN();
```

in the end, we could do even something like this:

```
diagramExportService
  .exportDiagram()
  .asXML()
  .asSVG()
  .asBPMN();

File: src/modules/bpmn-io/services/diagram.export.service.ts (hunk 6 above)
Comment (robpal94) [reply src/modules/bpmn-io/services/diagram.export.service.ts]: The power of two ;) 

Should I really make an own variable for this? I thought that the name `svgWidth**Squared**` would describe this enough.

File: src/modules/bpmn-io/services/diagram.export.service.ts [hunk 12]
```diff
@@ -0,0 +1,193 @@
...
+import * as beautify from 'xml-beautifier';
+import {
+  IBpmnModeler,
+  IDiagramExportService,
+  IProcessDefEntity,
+} from '../../../contracts';
+import {NotificationType} from '../../../contracts/notification/constants';
+import {NotificationService} from '../../notification/notification.service';
+
+export class DiagramExportService implements IDiagramExportService {
```
cmg-dev: I would seriously like an interface like this:

```ts
diagramExportService
  .exportDiagram()
  .asXML();

diagramExportService
  .exportDiagram()
  .asSVG();

diagramExportService
  .exportDiagram()
  .asBPMN();
```

in the end, we could do even something like this:

```ts
diagramExportService
  .exportDiagram()
  .asXML()
  .asSVG()
  .asBPMN();
```

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
+import * as download from 'downloadjs';
+import * as beautify from 'xml-beautifier';
+import {
```
cmg-dev: newline above...

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
...
+  IBpmnModeler,
+  IDiagramExportService,
+  IProcessDefEntity,
+} from '../../../contracts';
+import {NotificationType} from '../../../contracts/notification/constants';
+import {NotificationService} from '../../notification/notification.service';
+
+export class DiagramExportService implements IDiagramExportService {
+
+  private _modeler: IBpmnModeler;
```
cmg-dev: Why does this have to know about the modeler?

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
...
+  private _notificationService: NotificationService;
```
cmg-dev: This does not even need to know about the notification service.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  constructor(modeler: IBpmnModeler, notificationService: NotificationService) {
+    this._modeler = modeler;
+    this._notificationService = notificationService;
```
cmg-dev: The ONLY purpose of this is to export the diagrams into certain formats. 

This does not need to have anything to do with notification and the modeler.

src/modules/bpmn-io/services/diagram.export.service.ts
cmg-dev: If anybody will introduce something like this:

```ts
two: number = 2;
const svgWidthSquared: number = Math.pow(svgWidth, two);
```

I will seriously doubt this.

My preferred approach:

```ts
let square = number => number * number;

const svgWidthSquared: number = square(svgWidth);
```

This will work EVERYWHERE with no extra deps! 🤯mindblown.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
...
+        `An error occurred while processing the image for exporting.`);
+    }
+  }
+
+  /**
+   * Returns the current process model as a bpmn - xml.
+   *
+   * @returns The current process model as a bpmn - xml string.
+   */
+  public getXML(): Promise<string> {
```
cmg-dev: this is not needed here.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -0,0 +1,228 @@
...
+      });
+    });
+  }
+
+  /**
+   * Returns the current process model as a svg.
+   *
+   * @returns The current process model as a svg string.
+   */
+  public getSVG(): Promise<string> {
```
cmg-dev: also not needed here.

cmg-dev: I do not want to add the knowledge about the modeler to this exporter.

src/modules/bpmn-io/services/diagram.export.service.ts
cmg-dev: We only need the export service in the moment of pressing the button to export.

I certainly can live with, that we make `new` everytime that the button gets pressed.

So we can insert the XML and the SVG into the constructor.

src/modules/bpmn-io/services/diagram.export.service.ts
robpal94: Because in order to optain the SVG file, we have to call `modeller...getSVG()`. This method is needed by the exporting service because in order to export something, we have to acquire an SVG of the current process model somewhere.

src/modules/bpmn-io/services/diagram.export.service.ts
robpal94: The idea behind this was to show a notification, if the Diagram export failed. But yeah, I can move the error handling to `bpmn.io` since its the only place, where the export service is currently used yet.

robpal94:
_Just an **idea**_ 

What if we thread the Services like the `NotificationService` and make it injectable through the Aurelia Dependency Injector? So, like we do in the NotificationService, we could inject the Service to every component that needs this service.

src/modules/bpmn-io/bpmn-io.ts
robpal94: Then we have to write the following every time, we want to gather the current Process as an XML: 
```
this.modeler.saveXML({}, (error: Error, result: string) => {
        if (error) {
          this._notificationService.showNotification(NotificationType.ERROR, `Error while displaying the XML.`);
        } else {
          this.xml = result;
        }
      });
```
We could wrap this in an own method of the `bpmn-io` class but since we have a reference to the `exportService` anyway, it would be kinda redundant.

src/contracts/print/IDiagramPrintService.ts
```diff
@@ -0,0 +1,3 @@
+export interface IDiagramPrintService {
+  printDiagram(svgContent?: string): Promise<void>;
```
Vyperus: Please make this parameter not optional.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -1,14 +1,28 @@
 import * as bundle from '@process-engine/bpmn-js-custom-bundle';
 import {EventAggregator, Subscription} from 'aurelia-event-aggregator';
-import {bindable, inject, observable} from 'aurelia-framework';
+
```
Vyperus: Uneccesary blank line.

PR #527, opened 2018-07-06
♻️Refactor Services / Add fluent Interface to the export service

robpal94:
## What did you change?
This PR refactors the Export- and Print Services. 

The Export Service gets a kind of _fluent_ Interface. You can now use this Service like so: 

```
const exporter: IDiagramExportService = new DiagramExportService();

// Prettify and output an xml file from the modeller to a bpmn - diagram file
exporter.loadXML(this.xml)
    .asBpmn()
    .export('diagram.bpmn');

// Export the current SVG as an Image
exporter.loadSVG(this.svg);
    .asPNG()
    .export('diagram.png');
```

You also can initialize the exporter with an xml or svg:
```
const exporter: IDiagramExportService = new DiagramExportService(this.svg);

exporter.asSVG().export();
```
## PR-Checklist


src/contracts/export/IDiagramExportService.ts
```diff
@@ -1,8 +1,10 @@
 import {IProcessDefEntity} from '../processengine';
```
alexanderkasten: Remove unused import

robpal94:
@cmg-dev we need to discuss, what we need to add to the services. Maybe we can do this in another PR and merge this one, since this is more a pr meant for cleaning up the implemented services.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -4,32 +4,74 @@ import {IDiagramExportService} from '../../../contracts';
 
 export class DiagramExportService implements IDiagramExportService {
 
+  private _xml: string;
+  private _svg: string;
```
LeoTT: I don't see how adding state here is better.

Imo it just makes implementation more complicated & usage more confusing.
I don't see a use-case, where I would like to say 'service, just export whatever you have set for export right now'. Because who sets it(if not me) and why should I trust them?
It also introduces the wrong usage where you don't set the _xml and call exportBPMN without params.

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -4,32 +4,74 @@ import {IDiagramExportService} from '../../../contracts';
...
+      if (xml !== undefined && xml !== null) {
+        return xml;
+      }
+
+      if (this._xml !== undefined && this._xml !== null) {
+        return this._xml;
+      }
+
+      Promise.reject(`No XML defined in the exporter`);
+    })();
```
LeoTT: I like the idea of IIFE's, but in most cases I feel like the disadvantages outweigh the pros. Becomes slightly confusing because you have to differentiate which `return` statements belong to which scope.
```ts
const unformattedXML = xml ? xml : this.xml;
if (unformattedXML === undefined) {
   return Promise.reject....
}
```
Also, both xml and this._xml are in your hands. Therefore I would recommend not using null. If the user feeds null into this function, his fault. After all the type is not `exportBPMN(xml?: string | null)`.

src/modules/bpmn-io/services/diagram.export.service.ts
robpal94: I did this as a preparation for the eventually implemented _fluentAPI_ for this Service. I also thought that this may come handy in some situations.

src/modules/bpmn-io/services/diagram.export.service.ts
robpal94: With IIFE's, I can work around using `let` in this case. Should I change this to `let` and no IIFE instead?

src/modules/bpmn-io/services/diagram.export.service.ts
LeoTT: I don't think I understand that. There might be a vision for the future here that I don't see yet.
Is this the desired result?
```ts
diagramExportService.updateXML(xml).exportBPMN();
```

robpal94: Yes, kinda. @cmg-dev described that in this pr: https://github.com/process-engine/bpmn-studio/pull/523

> I would seriously like an interface like this:

```ts
diagramExportService
  .exportDiagram()
  .asXML();

diagramExportService
  .exportDiagram()
  .asSVG();

diagramExportService
  .exportDiagram()
  .asBPMN();
```

> in the end, we could do even something like this:

```ts
diagramExportService
  .exportDiagram()
  .asXML()
  .asSVG()
  .asBPMN();
```

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
+    return this;
+  }
+
+  public asSVG(): IExporter {
+    const currentStateIsNotSet: boolean = (this._currentState === undefined || this._currentState === null);
+    if (currentStateIsNotSet) {
+      throw new Error('No SVG file to export loaded');
+    }
+
+    // this._enqueuedExports.push(this._svgExporter);
```
alexanderkasten: Why is this commented out?

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
+
+  public asJPEG(): IExporter {
+    const currentStateIsNotSet: boolean = (this._currentState === undefined || this._currentState === null);
+    if (currentStateIsNotSet) {
+      throw new Error('No SVG file to convert and exporting defined.');
+    }
+
+    this._enqueuedExports.push(this._jpegExporter);
+    this._currentMimeType = 'image/jpeg';
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
+
+  public asPNG(): IExporter {
+    const currentStateIsNotSet: boolean = (this._currentState === undefined || this._currentState === null);
+    if (currentStateIsNotSet) {
+      throw new Error('No SVG file to convert and exporting defined.');
+    }
+
+    this._enqueuedExports.push(this._pngExporter);
+    this._currentMimeType = 'image/png';
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
+
+  public asSVG(): IExporter {
+    const currentStateIsNotSet: boolean = (this._currentState === undefined || this._currentState === null);
+    if (currentStateIsNotSet) {
+      throw new Error('No SVG file to export loaded');
+    }
+
+    // this._enqueuedExports.push(this._svgExporter);
+    this._currentMimeType = 'image/svg+xml';
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
-    return imageURL;
+  public asBpmn(): IExporter {
+    const currentStateIsNotSet: boolean = (this._currentState === undefined || this._currentState === null);
+    if (currentStateIsNotSet) {
+      throw new Error('No XML file to export loaded');
+    }
+
+    this._enqueuedExports.push(this._bpmnExporter);
+    this._currentMimeType = 'application/bpmn20-xml';
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
+   * TODO: Maybe we can do some kind of validation here to make sure, that the
+   * user who use this service really pass a valid svg here.
+   *
+   * @param svgContent updated svg file.
    */
-  public async exportPNG(svg: string): Promise<string> {
-    const svgToConvert: string = svg;
+  public loadSVG(svgContent: string): IExportSVG {
+    this._currentState = svgContent;
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
```diff
@@ -1,50 +1,157 @@
...
-   * TODO: This method is currently completely pointless. In a new PR, this
-   * will have the ability, to either return the svg delivered in the parameter
-   * or from the instance.
+   * @param xmlContent updated xml file.
    */
-  public async exportSVG(svg: string): Promise<string> {
-    return svg;
+  public loadXML(xmlContent: string): IExportXML {
+    this._currentState = xmlContent;
+    return this;
```
alexanderkasten: newline before return statement

src/modules/bpmn-io/services/diagram.export.service.ts
robpal94: Because its currently not necessary. Until we discuss the point of the `asSVG()` Method, we don't need to queue the call to `this._svgExporter()`.

src/modules/bpmn-io/services/diagram.print.service.ts
```diff
@@ -6,16 +6,28 @@ export class DiagramPrintService implements IDiagramPrintService {
...
+   *
+   * @param [svg] SVG content that should be printed
+   * @throws Error if no source SVG was defined.
    */
-  public async printDiagram(svg: string): Promise<void> {
-    const svgToPrint: string = svg || this._svg;
+  public async printDiagram(svg?: string): Promise<void> {
+    let svgToPrint: string;
+
+    if (svg !== undefined && svg !== null) {
```
LeoTT: Why would `svg` be `null`?
Calling `printDiagram(null)` seems like a weird idea to me and I think we should not encourage it.

If there ever is a case where you want to allow `null`, consider using `... | null` in the type.

src/contracts/exportService/IDiagramExportService.ts
```diff
@@ -1,16 +1,6 @@
-import {IExportSVG, IExportXML} from './index';
-export interface IDiagramExportService {
-  /**
-   * Load a new XML into the service.
-   *
-   * @param xmlContent Content of the xml, that should be loaded.
-   */
-  loadXML(xmlContent: string): IExportXML;
+import { ISvgConvertService, IXmlConvertService } from './index';
```
SteffenKn: Remove Whitespace

src/contracts/exportService/ISvgConvertService.ts
```diff
@@ -0,0 +1,7 @@
+import { IExportService } from './index';
```
SteffenKn: Remove Whitespace

src/modules/bpmn-io/services/diagram.svg.converter.ts
```diff
@@ -0,0 +1,168 @@
+import {IExportService, ISvgConvertService} from '../../../contracts/index';
+import { ExportService } from './export.service';
```
SteffenKn: Remove Whitespace

src/modules/bpmn-io/services/diagram.svg.converter.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  constructor(svgContent: string) {
+    this._svgContent = svgContent;
+  }
+
+  public asPNG(): IExportService {
+    const pngExporterPromise: Promise<string> = this._pngExporter();
+    this._enqueuedPromises.push(pngExporterPromise);
+    const mimeType: string = 'image/png';
+    return new ExportService(mimeType, this._enqueuedPromises);
```
SteffenKn: Imo this improves readability:

```
    const pngExporterPromise: Promise<string> = this._pngExporter();
    const mimeType: string = 'image/png';

    this._enqueuedPromises.push(pngExporterPromise);

    return new ExportService(mimeType, this._enqueuedPromises);
```

src/modules/bpmn-io/services/diagram.svg.converter.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  public asSVG(): IExportService {
+    const svgExporterPromise: Promise<string> = new Promise((resolve: Function): void => {
+      resolve(this._svgContent);
+    });
+    this._enqueuedPromises.push(svgExporterPromise);
+    const mimeType: string = 'image/svg+xml';
+    return new ExportService(mimeType, this._enqueuedPromises);
```
SteffenKn: Imo this improves readability:

```
    const mimeType: string = 'image/svg+xml';
    const svgExporterPromise: Promise<string> = new Promise((resolve: Function): void => {
      resolve(this._svgContent);
    });

    this._enqueuedPromises.push(svgExporterPromise);

    return new ExportService(mimeType, this._enqueuedPromises);
```

src/modules/bpmn-io/services/diagram.svg.converter.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+    /**
+   * Exports the current diagram as a PNG image.
+   */
+  private _pngExporter = async(): Promise<string> => {
+    return await this._generateImageFromSVG('png', this._svgContent);
```
SteffenKn: To the best of my knowledge, `return await` doesn't make any sense in this case.

src/modules/bpmn-io/services/diagram.svg.converter.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Exports the current diagram as a jpeg image.
+   */
+  private _jpegExporter = async(): Promise<string> => {
+    return await this._generateImageFromSVG('png', this._svgContent);
```
SteffenKn: To the best of my knowledge, `return await` doesn't make any sense in this case.

src/modules/bpmn-io/services/diagram.xml.converter.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  constructor(xmlContent: string) {
+    this._xmlContent = xmlContent;
+  }
+
+  public asBpmn(): IExportService {
+    const formatterPromise: Promise<string> = this._bpmnExporter();
+    this._enqueuedPromises.push(formatterPromise);
+    const mimeType: string = 'application/bpmn20-xml';
+    return new ExportService(mimeType, this._enqueuedPromises);
```
SteffenKn: Imo this improves readability:
```
    const formatterPromise: Promise<string> = this._bpmnExporter();
    this._enqueuedPromises.push(formatterPromise);
    const mimeType: string = 'application/bpmn20-xml';
    return new ExportService(mimeType, this._enqueuedPromises);
```

src/modules/bpmn-io/services/diagram.xml.converter.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Formats the current loaded xml.
+   */
+  private _bpmnExporter = async(): Promise<string> => {
+    const formatterPromise: Promise<string> = new Promise((resolve: Function): void => {
+      const formattedXml: string = beautify(this._xmlContent);
+      resolve(formattedXml);
+    });
+    return formatterPromise;
```
SteffenKn: Add some spacing:
```
    const formatterPromise: Promise<string> = new Promise((resolve: Function): void => {
      const formattedXml: string = beautify(this._xmlContent);

      resolve(formattedXml);
    });

    return formatterPromise;
```

src/modules/bpmn-io/services/export.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  public async export(filename: string): Promise<void> {
+
+    let contentToExport: string;
+
+    /*
+     * Wait, until all queued functions are executed
+     */
+    for (const currentPromise of this._enqueuedPromises) {
+      contentToExport = await currentPromise;
+    }
```
SteffenKn: I don't get this.

For every enqueued promise you overwrite contentToExport. So you only export the last promise?

I think you wanted to do something like this:
```
const contentsToExport: string = await Promise.all(this._enqueuedPromises);

for(const content of contentsToExport) {    
  this._exportDiagramRepository.exportDiagram(content, filename, this._currentMimeType);
}

this._enqueuedPromises = [];
```

src/modules/bpmn-io/services/diagram.svg.converter.ts
alexanderkasten: @SteffenKn Yes, just return is enough

src/modules/bpmn-io/services/diagram.svg.converter.ts
alexanderkasten: @SteffenKn Yes, just return is enough

src/modules/bpmn-io/services/export.service.ts
robpal94: Im not sure, if `Promise.all()` keeps the order of the passed list. This is necessary because the services methods rely on the ordered execution (there is a more _functional_ approach using `array.reduce()`, bit its very very ugly to read imo). 

Generally, Im only interested in the value of the last resolved promise but I don't know, how to obtain only the last resolved value of a list which contains promises.

SteffenKn: Wait you await every single promise to just get the value of the last one?
That doesn't sound very efficient.

Why is _enqueuedPromises `Array<Promise<string>>? You could just use `<Promise<string>>` and only store the last value.

When you want to get the last element of an array you can use array.pop().
This also removes the element from the array, but this should be ok since you clear the array anyways.

Issue #469, opened 2018-06-05
Enable no-any rule again


LeoTT:
#### Describe your issue

At the moment the `no-any` rule is disabled and the `typedef` rule is enabled.
This forces developers to use `any` where no type is better.
EDIT: I don't mean to encourage using no type, but if in doubt it is the lesser evil.

#### 1. Because you lose the ability to express that you actually mean *anything* as a type
```ts
// is this function able to log anything? or is the expected type just unknown? who knows?
function log (anything: any): void {
...
}
```

#### 2. Because you kill type inference
```ts
// this is correct but annoying
const options: {
    host: string,
    port: number,
    userName: string,
    password: string,
} = {
    host: 'localhost',
    port: 8912,
    userName: 'admin',
    password: 'admin',
}

// this looks a bit better, but type information lost -> properties no longer known
const options: { 
 [index: string] : string | number,
} = {
    host: 'localhost',
    port: 8912,
    userName: 'admin',
    password: 'admin',
}

// absolute typeloss -> this is currently encouraged by tslint settings
const options: any = {
    host: 'localhost',
    port: 8912,
    userName: 'admin',
    password: 'admin',
}

// type inference gets type
const options = { // tslint:disable-line:typedef
    host: 'localhost',
    port: 8912,
    userName: 'admin',
    password: 'admin',
}
```
I prefer the last example. This does not mean I would get rid of the `typedef`rule, I just think it's reasonable to disable tslint for it sometimes.

#### Possible solution

Remove the expection from `tslint.json`.

Replace occurences of `any` with actual type or (if not possible) with no type + `tslint-disable-line`.

Here is a list of the occurences(59):
https://github.com/process-engine/bpmn-studio/blob/3aa9d7177391cde4e1b14162bf0f92dee26e0f3e/src/app.ts#L7

https://github.com/process-engine/bpmn-studio/blob/3aa9d7177391cde4e1b14162bf0f92dee26e0f3e/src/contracts/bpmnmodeler/bpmnElements/IModdleElement.ts#L10

https://github.com/process-engine/bpmn-studio/blob/3aa9d7177391cde4e1b14162bf0f92dee26e0f3e/src/contracts/bpmnmodeler/IBpmnModdle.ts#L6

https://github.com/process-engine/bpmn-studio/blob/3aa9d7177391cde4e1b14162bf0f92dee26e0f3e/src/contracts/bpmnmodeler/IBpmnModeler.ts#L4

[...]

PR #573, opened 2018-07-17
[WIP] login using identity server (browser only)

5ebastianMeier:
## What did you change?

* added open-id-connect support to be able to login to IdentityServer
   * switched routing to push state to allow the library to work properly
* added a `NewAuthenticationService` that is meant to replace the old `AuthenticationService`
   * the full replacement will take place in a further PR, because that involves other parts of the application to switch to the new object model
* modified the `IIdentity` interface to orient reflect the properties from an identity resource retrieved by querying the `connect/userinfo` route on the `IdentityServer`
* added a modal dialog to choose a start event before starting a process instance
* added a new section to the config panel to change the authority (`IdentityServer`) url at runtime
* changed code comments that mimiced regions to actual regions

> Describe how others can test your changes (you can remove this section for typo fixes etc.)

## PR-Checklist

- [ ] I've merged the `develop` branch into my branch before finishing this PR.
- [ ] I've mentioned all **PRs, which relate to this one**


src/open-id-connect-configuration.ts
```diff
@@ -0,0 +1,31 @@
...
+  unauthorizedRedirectRoute: '',
+  logLevel: 0,
+  userManagerSettings: {
+    accessTokenExpiringNotificationTime: 1,
+    authority: environment.openIdConnect.authority,
+    automaticSilentRenew: true,
+    monitorSession: true,
+    checkSessionInterval: 2000,
+    client_id: 'bpmn_studio',
+    // client_id: 'implicit',
```
cmg-dev: RM Comment.

src/contracts/authentication/IIdentity.ts
```diff
@@ -1,5 +1,7 @@
 export interface IIdentity {
   id: string;
   name: string;
+  given_name: string;
+  family_name: string;
```
LeoTT: Later on you make the following check:
```
if (!this.identity.given_name || !this.identity.family_name) {
...
```
Should `given_name` and `family_name` be optional?

Also, the interface already exists in core contracts: https://github.com/essential-projects/core_contracts/blob/78c00976c0755e792ca01de468a0f2d308ef6d3d/src/iam/iidentity.ts#L1 (without those two additions though). 
If it is the same thing I would rather add `export {IIdentity} from '@essential-projects/core_contracts';` in the index.ts of authentication.

src/modules/config-panel/config-panel.ts
```diff
@@ -16,32 +18,40 @@ export class ConfigPanel {
   private _eventAggregator: EventAggregator;
   private _authenticationService: IAuthenticationService;
   private _subscriptions: Array<Subscription>;
+  // We use any here, because we need to call private members (see below)
+  private _openIdConnect: OpenIdConnect | any;
 
-  public config: any = environment.bpmnStudioClient;
-  public isLoggedIn: boolean;
+  public config: any = environment;
```
LeoTT: type is `typeof environment`

src/contracts/authentication/IIdentity.ts
5ebastianMeier: The one in `core_contracts` is obsolete and the BPMN-Studio can define its own interface for that, because it resembles the identity resources (see [IdentityServer Identity Resources](http://docs.identityserver.io/en/release/topics/resources.html#defining-identity-resources)) the BPMN-Studio uses.

Those informations might differ between multiple applications so that I'm fine with each application defining its own interface for this.

You're right that these fields should be optional.

src/modules/config-panel/config-panel.ts
5ebastianMeier: You're right that this is better than `any`. But this also a prime example of where type inference would spare us doing that kind of redundant type annotations.

LeoTT: Right, made an issue about that #469.
What I dislike about the current situation is that `any`(arguably the worst decision) is the easiest way to satisfy the tslint settings.

src/modules/config-panel/config-panel.ts
```diff
@@ -16,32 +18,42 @@ export class ConfigPanel {
...
     this._router = router;
     this._bpmnStudioClient = bpmnStudioClient;
-    this.config.processEngineRoute = environment.bpmnStudioClient.baseRoute;
     this._notificationService = notificationService;
     this._eventAggregator = eventAggregator;
     this._authenticationService = authenticationService;
+    this._openIdConnect = openIdConnect;
+  }
+
+  private _initializeConfig(): void {
```
alexanderkasten: This method is never used

src/modules/config-panel/config-panel.ts
```diff
@@ -16,32 +18,42 @@ export class ConfigPanel {
...
+    this.config.bpmnStudioClient.baseRoute = environment.bpmnStudioClient.baseRoute;
+    this.config.openIdConnect.authority = environment.openIdConnect.authority;
   }
 
   public attached(): void {
-    this.isLoggedIn = this._authenticationService.hasToken();
+    this.baseRoute = this.config.bpmnStudioClient.baseRoute;
+    this.isLoggedInToProcessEngine = this._authenticationService.hasToken();
```
alexanderkasten: We are using the old authentication service here, shouldn't we use your implemented `NewAuthenticationService` instead?

src/modules/processdef-list/processdef-list.html
```diff
@@ -90,7 +90,6 @@ <h1 class="modal-dialog__headline">Diagram already exists!</h1>
           <tr repeat.for="process of processes" id="processDef">
             <td class="process-name-table-entry" title="${process.name}">${process.name}</td>
             <td class="process-buttons-table-entry">
-              <button click.delegate="startProcess(process.id)" class="btn btn-success">Start</button>
```
alexanderkasten: Why is this button and its method removed?

src/modules/processdef-start/processdef-start.ts
```diff
@@ -87,7 +80,4 @@ export class ProcessDefStart {
     }
   }
 
-  private _startProcess(): void {
-    this._bpmnStudioClient.startProcessByKey(this.process.key);
```
alexanderkasten: When removing this method, the `CreateProcessDef` button above the table on the processdef-list is been obsolet.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -40,18 +52,30 @@ export class ProcessDefDetail {
   private _diagramIsInvalid: boolean = false;
   // Used to control the modal view; shows the modal view for pressing the play button.
   private _startButtonPressed: boolean = false;
+  private _authenticationService: IAuthenticationService;
+  private _managementApiClient: IManagementApiService;
+
+  public dropdownArrow: HTMLElement;
```
alexanderkasten: This variable is never used.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -40,18 +52,30 @@ export class ProcessDefDetail {
...
+  private dropdownIsCollapsed: boolean = true;
```
alexanderkasten: This variable is never used.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -321,15 +332,27 @@ export class ProcessDefDetail {
...
+    */
     try {
       await this._updateProcessStartEvents();
     } catch (error) {
       this.
         _notificationService
         .showNotification(
           NotificationType.ERROR,
-          error.message,
+          `Could not load the processes start Events: ${error.message}`,
```
alexanderkasten: `StartEvents` / `start events`

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -40,18 +52,27 @@ export class ProcessDefDetail {
   private _diagramIsInvalid: boolean = false;
   // Used to control the modal view; shows the modal view for pressing the play button.
   private _startButtonPressed: boolean = false;
+  private _authenticationService: IAuthenticationService;
+  private _managementApiClient: IManagementApiService;
+
+  public processesStartEvents: Array<Event> = [];
```
alexanderkasten: Move the public variables to the top

src/modules/config-panel/config-panel.ts
5ebastianMeier: No.

The `NewAuthenticationService` is only used in the few places where we need the new behavior.
For everything still related to the old process engine backend stack, we still use the old implementation to not mix things up by accident.

In another PR we will refactor all the leftovers that still use the old implementation, but that is independent of the scope of this PR.

src/modules/processdef-start/processdef-start.ts
5ebastianMeier: Until now this feature used a process on the backend to create another process.

Now that we use a more file-centric approach I'd like to refactor this, so that local files are created - rather than doing the roundtrip on the backend.

@cmg-dev @roblen45 do you want to see this fixed, or can we note a known issue for this and implement it like mentioned above?

5ebastianMeier:
@alexanderkasten can't reproduce your bug from https://github.com/process-engine/bpmn-studio/pull/573#pullrequestreview-138547497 - please make sure you got a fresh install and try again

src/contracts/authentication/IAuthenticationService.ts
```diff
@@ -1,9 +1,10 @@
 import {IIdentity} from './IIdentity';
 
 export interface IAuthenticationService {
-  login(username: string, password: string): Promise<IIdentity>;
+  // TODO: remove old implementations that still use username and password
+  login(username?: string, password?: string): Promise<void>;
   logout(): Promise<void>;
-  getToken(): string;
+  getAccessToken(): string;
   hasToken(): boolean;
```
Paulomart: Rename this to `hasAccessToken()`

src/modules/authentication/authentication.service.ts
```diff
@@ -44,15 +45,15 @@ export class AuthenticationService implements IAuthenticationService {
...
     return this.tokenRepository.getToken();
   }
 
-  public getIdentity(): IIdentity {
-    return this.tokenRepository.getIdentity();
+  public getIdentity(): Promise<IIdentity> {
+    return Promise.resolve(this.tokenRepository.getIdentity() as IIdentity);
   }
 
-  public async login(username: string, password: string): Promise<IIdentity> {
```
Paulomart: Please improve the interface for this method or change the return type to void.

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+import {OpenIdConnect} from 'aurelia-open-id-connect';
+import {User} from 'oidc-client';
+
+import {AuthenticationStateEvent, IAuthenticationService, IIdentity} from '../../contracts/index';
+import environment from './../../environment';
+
+const UNAUTHORIZED_STATUS_CODE: number = 401;
+
+@inject(EventAggregator, OpenIdConnect)
+export class NewAuthenticationService implements IAuthenticationService {
```
Paulomart: IMO: This is a **OpenId**AuthenticationService and should be named so

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  private _openIdConnect: OpenIdConnect;
+  private _user: User;
+
+  constructor(eventAggregator: EventAggregator, openIdConnect: OpenIdConnect) {
+    this._eventAggregator = eventAggregator;
+    this._openIdConnect = openIdConnect;
+    this.initialize();
+  }
+
+  public async initialize(): Promise<void> {
```
Paulomart: can this be private?

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+export class NewAuthenticationService implements IAuthenticationService {
+
+  private _eventAggregator: EventAggregator;
+  private _openIdConnect: OpenIdConnect;
+  private _user: User;
+
+  constructor(eventAggregator: EventAggregator, openIdConnect: OpenIdConnect) {
+    this._eventAggregator = eventAggregator;
+    this._openIdConnect = openIdConnect;
+    this.initialize();
```
Paulomart: async function is not awaited

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  public async initialize(): Promise<void> {
+    this._user = await this._openIdConnect.getUser();
+  }
+
+  public async login(): Promise<void> {
```
Paulomart: I dont see the point for changing the interface to void here. You could easily just return the Identity.

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        this._openIdConnect.logout();
+      } catch (error) {
+        reject(error);
+      }
+
+    });
+  }
+
+  public getAccessToken(): string {
+    if (!this._user) {
```
Paulomart: use `undefined` and `null` checks here.

src/modules/authentication/new_authentication.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  public async getIdentity(): Promise<IIdentity> {
+
+    const accessToken: string = this.getAccessToken();
+
+    if (!accessToken) {
+      return null;
+    }
+
+    const request: Request = new Request(`${environment.openIdConnect.authority}/connect/userinfo`, {
```
Paulomart: This should be moved to a repository

src/modules/management-api_client/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const clientService: ManagementApiClientService = createManagementApiClient();
+
+  config.container.registerInstance('ManagementApiClientService', clientService);
+}
+
+function createManagementApiClient(): ManagementApiClientService {
+
+  const httpClient: any = {
+    post: async(url: string, payload: any, headers: any): Promise<any> => {
```
Paulomart: Please create a Repository for the http client.

Please create typing for the http client, the return type and headers

src/modules/processdef-detail/processdef-detail.scss
```diff
@@ -18,26 +18,21 @@
   display: none;
 }
 
-.process-details-dropdown-menu {
+/*
+ * TODO: Find out where this class is actually used.
+ */
+/*.process-start-event-modal-dialog__dropdown-menu {
```
Paulomart: If this class not referenced you can remove it.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -22,11 +26,22 @@ interface RouteParameters {
...
+  'NewAuthenticationService',
+  'ManagementApiClientService')
 export class ProcessDefDetail {
 
   public bpmnio: BpmnIo;
   public process: IProcessDefEntity;
+  public showModal: boolean = false;
+
+  public processesStartEvents: Array<Event> = [];
+  public dropdownMenu: HTMLSelectElement;
```
Paulomart: Please be aware that we are currently trying to remove direct references on html. See #455 of more details.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -217,6 +243,51 @@ export class ProcessDefDetail {
...
+     * Create a promise which displays the modal and resolves, if the user
+     * clicks on the buttons.
+     */
+    const returnPromise: Promise<string> = new Promise((resolve: Function, reject: Function): void => {
+      const cancelButton: HTMLElement = document.getElementById('cancelStartEventSelection');
+      const startProcessButton: HTMLElement = document.getElementById('startProcessWithSelectedStartEvent');
+
+      cancelButton.addEventListener('click', () => {
+        this.showModal = false;
+        resolve('');
```
Paulomart: Why not reject with null and add `Promise<string | null>` as type

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -239,12 +310,64 @@ export class ProcessDefDetail {
...
+    try {
+      await this._updateProcessStartEvents();
+    } catch (error) {
+      this
+        ._notificationService
+        .showNotification(
+          NotificationType.ERROR,
+          `Error while obtaining the StartEvents which belongs to the Process: ${error.message}`,
+        );
+      throw Error(error);
```
Paulomart: Why not just `throw error;` ?

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -239,12 +310,64 @@ export class ProcessDefDetail {
...
+    /*
+     * If the process has only one defined start event, we dont need to
+     * ask the user, from which start event he wants to start the process since
+     * we always can pick the normal start event.
+     */
+    const selectedStartEvent: string = (this.processesStartEvents.length > 1)
+                                        ? await this.showModalDialogAndAwaitAnswer()
+                                        : this.processesStartEvents[0].id;
+
+    if (selectedStartEvent === '') {
```
Paulomart: again i would prefer null here.

src/modules/user-login/user-login.html
```diff
@@ -1,11 +1,12 @@
...
-      <span>${identity ? identity.name : 'Guest'} </span>
-      <i class="fas fa-caret-down"></i>
-    </a>
-    <div ref="dropdown" class="dropdown-menu dropdown-menu-right user-login-dropdown">
+    <button show.bind="!isLoggedIn" click.delegate="login()">login</button>
+    <div show.bind="isLoggedIn">
+      <span>${username}</span>
+      <button click.delegate="logout()">logout</button>
+    </div>
+    <!-- <div ref="dropdown" class="dropdown-menu dropdown-menu-right user-login-dropdown">
```
Paulomart: Please remove code that is commented out.

src/open-id-connect-configuration.ts
```diff
@@ -0,0 +1,30 @@
+import {OpenIdConnectConfiguration} from 'aurelia-open-id-connect';
+import {WebStorageStateStore} from 'oidc-client';
+import environment from './environment';
+
+export const oidcConfig: OpenIdConnectConfiguration = {
```
Paulomart: Why is this _config_ in the src folder?

tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "moduleResolution": "node",
-    "lib": ["es2017", "dom"]
+    "lib": ["esnext", "dom"]
```
Paulomart: I dont think this is intended?

src/modules/processdef-start/processdef-start.ts
cmg-dev: File-centric is the way to go.

The old process is rather old-fashioned. We will be able to deploy diagrams through the management api.
This will be our "new" workflow:

> Create local diagram -> Design it -> Deploy to ProcessEngine -> Run it.

Deploy and Run use functions from the Management API.

Paulomart:
@5ebastianMeier Please resolve merge conflicts and rebase on develop.

src/contracts/authentication/IAuthenticationService.ts
5ebastianMeier: I think the whole method can be removed safely in the following PR where we get rid of the old authentication service and can clean up the interface from the obsolete use cases

src/modules/authentication/authentication.service.ts
5ebastianMeier: The whole service can be removed safely in the following PR where we get rid of the old authentication service and can clean up the interface from the obsolete use cases

src/modules/authentication/new_authentication.service.ts
5ebastianMeier: This will replace the original `AuthenticationService` and I'd like to not mention `OpenId` as long as there are no other implementations we support.

Paulomart: Even if we had no other Auth Service, I woudl still prefer the OpenId name.

src/modules/authentication/authentication.service.ts
Paulomart: Okay.

src/modules/authentication/new_authentication.service.ts
Paulomart: This is still not awaited.

5ebastianMeier:
@SteffenKn this is due to push state being activated and usage of static links on routes including `#` instead of letting aurelia build the url - fixed it for now

src/modules/authentication/new_authentication.service.ts
ElRaptorus: I agree with @5ebastianMeier, putting the implementation in the name of the class only makes sense, if you have multiple implementations for something.
And even then, it is better to still have a single service with a plain name for the use case it covers and then use a repository for each implementation.
Those repositories can each carry the name of the technology they implement.

5ebastianMeier: Not for the service - a repository specific to `OpenId` could be discussed - but then again, I'd like to have a second use case before refactoring everything

Paulomart: okay then

src/modules/authentication/new_authentication.service.ts
ElRaptorus: What for? Using `!` covers both. 
Imho, checking both cases separately only makes sense if you want to handle them differently.

Paulomart: We still prefer `this._user === null || this._user === undefined` in the bpmn studio codestyle.

src/modules/authentication/new_authentication.service.ts
Paulomart: Please open an issue for that.

src/modules/management-api_client/index.ts
Paulomart: Please open an issue for that.

src/modules/processdef-detail/processdef-detail.ts
Paulomart: Please open an issue for that.

PR #589, opened 2018-07-19
🐛Fix save dialog when leaving an unsaved diagram

### LINKED ISSUE #542: 🐛 Process Always Contains Changes
#### Describe your issue

If you open this [process](https://github.com/process-engine/bpmn-studio/files/2187720/process.zip) in BPMN Studio, it will always contain changes.

grafik

- Open via file repo or import the process 
- Open the process via solution explorer
- Go to another process via solution explorer
- Dialog shows, even with no changes made!

#### Issue checklist

robpal94:
## What did you change?
This PR fixes an issue, where the _unsaved changes_ modal was shown, even if the diagram does not contain any actual changes. 

Also fixes #542 

* Load a diagram
* Make a change
* Revert the change
* Navigate back from the detail view

## PR-Checklist


robpal94:
Hmm, I thought the `diff` library where a bit smarter.

SteffenKn:
The only problem is that the diff result is equal when the diagrams are identical and when the diagrams are incomparable.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -209,18 +209,20 @@ export class BpmnIo {
...
+        const unformattedXml: string = currentXml.replace(whitespaceAndNewLineRegex, '');
+        const unformattedSaveXml: string = this.savedXml.replace(whitespaceAndNewLineRegex, '');
 
-        const savingNeeded: boolean = objectIsNotEmpty(this.xmlChanges._added)
-          || objectIsNotEmpty(this.xmlChanges._changed)
-          || objectIsNotEmpty(this.xmlChanges._layoutChanged)
-          || objectIsNotEmpty(this.xmlChanges._removed);
+        const diagramIsUnchanged: boolean = unformattedSaveXml === unformattedXml;
+
+        this._eventAggregator.publish(environment.events.diagramNeedsSaving, !diagramIsUnchanged);
```
SteffenKn: Imo this is easier to read:

```
const diagramIsChanged: boolean = unformattedSaveXml !== unformattedXml;

this._eventAggregator.publish(environment.events.diagramNeedsSaving, diagramIsChanged);
```

SteffenKn: Imo this is much easier to read:

```
const diagramIsChanged: boolean = unformattedSaveXml !== unformattedXml;

this._eventAggregator.publish(environment.events.diagramNeedsSaving, diagramIsChanged);
```

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -219,10 +219,8 @@ export class BpmnIo {
         const unformattedXml: string = currentXml.replace(whitespaceAndNewLineRegex, '');
         const unformattedSaveXml: string = this.savedXml.replace(whitespaceAndNewLineRegex, '');
 
-        const diagramIsUnchanged: boolean = unformattedSaveXml === unformattedXml;
-
-        this._eventAggregator.publish(environment.events.diagramNeedsSaving, !diagramIsUnchanged);
-
+        const diagramIsChanged: boolean = unformattedSaveXml !== unformattedXml;
+        this._eventAggregator.publish(environment.events.diagramNeedsSaving, diagramIsChanged);
```
SteffenKn: Please add some spacing.
```
const diagramIsChanged: boolean = unformattedSaveXml !== unformattedXml;

this._eventAggregator.publish(environment.events.diagramNeedsSaving, diagramIsChanged);
```

robpal94:
Using the generated XML as a reference seems to be pretty unreliable. Maybe, with the changes added to the new Version of the differ, [see here](https://github.com/bpmn-io/bpmn-js-differ/commit/c6dd4732180469a5e46a4bad691df6854185745e), to find out if the diagram contains unsaved changes.

aurelia_project/environments/dev.ts
```diff
@@ -20,6 +20,7 @@ export default {
     },
   },
   events: {
+    diagramNeedsSaving: 'diagramNeedsSaving',
```
Vyperus: Thats not a good event name. It sounds like a boolean. Please use something that describes what happened or should happen wenn this event is fired. "diagramChanged" would be a good name.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -20,6 +21,7 @@ import {NotificationService} from './../notification/notification.service';
 import {DiagramExportService, DiagramPrintService} from './services/index';
 
 import * as download from 'downloadjs';
+import { async } from '../../../node_modules/@types/q';
```
Vyperus: Remove whitespaces. I'm pretty sure there is a way to not use this type of notation ('../../../node_modules....').

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -99,8 +99,14 @@ export class ProcessDefDetail {
...
-      this._eventAggregator.subscribe(environment.events.diagramChange, () => {
-        this._diagramHasChanged = true;
+      this._eventAggregator.subscribe(environment.events.diagramNeedsSaving, (savingNeeded: boolean) => {
+        /*
+         * Using a tenary expression instead a direct assigment makes sure, that
+         * diagramHasChanges never contains undefined as a value.
+         */
+        this._diagramHasChanged = (savingNeeded === true)
+          ? (this._diagramHasChanged = true)
+          : (this._diagramHasChanged = false);
```
LeoTT: I don't get this. In the following code example `_diagramHasChanged` will be either false or true, never `undefined`.
```
this._diagramHasChanged = (savingNeeded === true);
```

I find the current code pretty confusing. Could not tell whether it works or not, had to test. `variableName = true` is a truthy value, `variableName = false` is a falsy value. The actual assignment gets executed and overwritten by the truthiness of the assignment(which is luckily the same). This seems like one of those riddles for a coding interview.

src/modules/bpmn-io/bpmn-io.ts
LeoTT: `import {async} from 'q'` would do the trick (although it would not only import the type but also the function and overwriting the keyword `async` seems dangerous).
Is this needed? Opened the file and seems to work just fine without this type.

aurelia_project/environments/dev.ts
robpal94: There already is an event named _diagramChanged_, but its used for another purpose.

src/modules/bpmn-io/bpmn-io.ts
robpal94: Seems like vscode auto imported this, since Im not using it anywhere. I will remove it.

src/modules/processdef-detail/processdef-detail.ts
robpal94: My original intention was to make sure, that the content of `_diagramHasChanged` is a boolean in every case and not something like `undefined` (since TypeScript has no way to check the type of this parameter, when the event is thrown, using `eventAggregator.publish`). 

The above check just simply assigns `false`, if the `savingNeeded` parameter is a value that would evaluate to false. Just to be a bit more _robust_. 

Sounds like an excuse, but keep in mind that im not that long into this JS/TS stuff. I will remove the statement.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -205,6 +206,22 @@ export class BpmnIo {
...
+        */
+        const whitespaceAndNewLineRegex: RegExp = /\r?\n|\r|\s/g;
+
+        const currentXml: string = await this.getXML();
+        const unformattedXml: string = currentXml.replace(whitespaceAndNewLineRegex, '');
+        const unformattedSaveXml: string = this.savedXml.replace(whitespaceAndNewLineRegex, '');
+
+        const diagramIsChanged: boolean = unformattedSaveXml !== unformattedXml;
+
+        this._eventAggregator.publish(environment.events.diagramNeedsSaving, diagramIsChanged);
```
LeoTT: Is there any value in publishing this event if `diagramIsChanged` is false?
Seems to me like you could just say that this event only fires if the diagram is changed (and then it would not need to carry that boolean anymore).

Makes it also more clear imo. If I'd subscribe to an event with that name I would expect that the diagram does need saving.

aurelia_project/environments/dev.ts
LeoTT: Don't have a good name but I would also think about renaming this.

Maybe `unsavedChangesOccured`?

src/modules/bpmn-io/bpmn-io.ts
robpal94: The thing is that the ProcessDef - Detail, which handles the saving modal, has to know, if there are unsaved changes present or not.

If I only would throw this event, when there are unsaved changes, and revert the changes with ctrl + z, the ProcessdefDetail would stiff think, that they are unsaved changes. 

tldr; The value is `false`, if the user reverts his changes using ctrl + z.

LeoTT: Are you certain that the parameter is needed? 
This is the current code:
```ts
// publishing:
...
this._eventAggregator.publish(environment.events.changesOccurred, diagramIsChanged); 
...
// subscribing:
...
this._eventAggregator.subscribe(environment.events.changesOccurred, (savingNeeded: boolean) => {
  this._diagramHasChanged = savingNeeded === true;
});
...
```
This is how I would refactor it:
```ts
// publishing:
...
if (diagramIsChanged) {
  this._eventAggregator.publish(environment.events.changesOccurred); 
}
...
// subscribing:
this._eventAggregator.subscribe(environment.events.changesOccurred, () => {
  this._diagramHasChanged = true;
});
...
```

I'm just looking at the code here, but seems to me like same effect.

robpal94: The parameter can be set to true **or** to false. 

If the user made some changes, that are **actual changes** (if the xml is updated), the event will be fired like so: 
```
this_eventAggregator.publish(..., true);
```

If the user undoes his changes, the event will be fired like so: 
```
this_eventAggregator.publish(..., false);
```

In the subscribing class, I **need** to know, if the parameter is true or false because I need to know if the changes result in a changed xml that has to be saved. 

Using an updated version of the `bpmn-differ` would make this way more easier.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -118,9 +118,9 @@ export class ProcessDefDetail {
 
       //#endregion
 
-      //#region General Event Subscritions
-      this._eventAggregator.subscribe(environment.events.diagramChange, () => {
-        this._diagramHasChanged = true;
+      //  General Event Subscritions {{{ //
+      this._eventAggregator.subscribe(environment.events.changesOccurred, (savingNeeded: boolean) => {
+        this._diagramHasChanged = savingNeeded === true;
```
LeoTT: I think I understand the parameter now.

You can use `this._diagramHasChanged = savingNeeded;` here.

Would also consider to rename `savingNeeded` to `differsFromOriginal` or something like that.

src/environment.ts
```diff
@@ -0,0 +1,76 @@
+const processEngineRoute: string = 'http://localhost:8000';
```
SteffenKn: Remove the environment.ts

robpal94: This will be done when rebasing onto the develop branch since the environement.ts was added to the gitignore file.

src/modules/processdef-detail/processdef-detail.ts
```diff
@@ -118,9 +118,9 @@ export class ProcessDefDetail {
 
       //#endregion
 
-      //#region General Event Subscritions
-      this._eventAggregator.subscribe(environment.events.diagramChange, () => {
-        this._diagramHasChanged = true;
+      //  General Event Subscritions {{{ //
```
cmg-dev: please revert this line. no need to change it.

robpal94: Wo what should I revert this line? Since I don't touched the region lines, this change was made during the rebase from develop.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -95,9 +96,9 @@ export class BpmnIo {
      */
     const handlerPriority: number = 1000;
 
-    this.modeler.on('commandStack.changed', () => {
+    this.modeler.on('commandStack.changed', async() => {
       this._eventAggregator.publish(environment.events.diagramChange);
-    }, handlerPriority);
+      }, handlerPriority);
```
SteffenKn: Wrong Indentation

cmg-dev:
What about this @robpal94 ?

Is this still a thing?

robpal94:
@cmg-dev this PR waits for an approval. It can be merged if it gets approved.

src/modules/bpmn-io/bpmn-io.ts
```diff
@@ -95,13 +96,12 @@ export class BpmnIo {
      */
     const handlerPriority: number = 1000;
 
-    this.modeler.on('commandStack.changed', () => {
+    this.modeler.on('commandStack.changed', async() => {
```
alexanderkasten: why adding async here?

PR #632, opened 2018-07-26
Refactor navigation

5ebastianMeier:
## What did you change?

* changed all `router.navigate()` calls to `router.navigateToRoute()`
* changed all `<a href="">` tags to `<a route-href="">`
* refactored former logic that checked if the route title matches to `isRouteActive` that uses the same name as `router.navigateToRoute()` and `<a route-href="">`

## PR-Checklist


package.json
```diff
@@ -82,7 +82,7 @@
   "dependencies": {
     "@essential-projects/errors_ts": "^1.3.0",
     "@essential-projects/http_contracts": "^2.0.2",
-    "@process-engine/skeleton-electron": "^5.0.0-pre3",
+    "@process-engine/skeleton-electron": "^5.0.0",
```
Vyperus: Update dependency since 5.0.0 is not working in the current state

src/modules/property-panel/indextabs/general/sections/call-activity/call-activity.ts
```diff
@@ -48,7 +48,9 @@ export class CallActivitySection implements ISection {
   }
 
   public navigateToCalledProcess(): void {
-    this._router.navigate(`/processdef/${this.selectedProcess.id}/detail`);
+    this._router.navigateToRoute('processdef-detail', {
+      processDefId: this.selectedProcess.id,
+    });
```
robpal94: **Just an Idea**
Maybe we can create an interface for the type

```
{
    processDefId: string
}
```

src/modules/process-solution-panel/process-solution-panel.ts
```diff
@@ -224,7 +224,9 @@ export class ProcessSolutionPanel {
 
   public async navigateToDiagramDetail(diagram: IDiagram): Promise<void> {
     this._eventAggregator.publish(environment.events.navBar.updateProcess, diagram);
-    this._router.navigateToRoute('diagram-detail', {diagramUri: diagram.uri});
+    this._router.navigateToRoute('diagram-detail', {
+      diagramUri: diagram.uri,
+    });
```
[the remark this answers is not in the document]
robpal94: This would also apply to the other calls to `router.navigateToRoute()` that using special route params.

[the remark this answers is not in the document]
5ebastianMeier: I'm not totally against that recommendation, but it's a whole lot of boilerplate we get for no added value.

The type annotation would have to be centralized and I doubt it would make the code any more readable.

src/modules/property-panel/indextabs/general/sections/call-activity/call-activity.ts
5ebastianMeier: `router.navigateToRoute` is not a generic function - so we would need to have an extra variable with extra type annotation, just for the type safety.

If we get more complex routes and revisit the navigation declaration, I'd be open to discussing a type safe approach - but imo currently the disadvantages outweigh the benefits

Issue #692, opened 2018-08-01
UI faulty in `Tasks waiting` 


LeoTT:
#### Describe your issue

<img width="916" alt="screen shot 2018-08-01 at 15 23 55" src="https://user-images.githubusercontent.com/21304781/43524262-200346a4-959f-11e8-8e26-a31c9baefc0a.png">

2 things: 
`Tasks waiting` window fills the full screen, probably within a wrong bootstrap container.
~~The table contains the userTaskId twice.~~
EDIT: Double user task column was fixed somewhere along the way.
#### Issue checklist

Please check the boxes in this list after submitting your Issue:

- [x] I've checked if this issue already exists
- [x] I've included all the information that i think is relevant
- [x] I've added logs and/or screenshots (if applicable)
- [x] I've mentioned PRs and issues that relate to this one

PR #767, opened 2018-08-07
💄 Add padding to tasklist

LeoTT:
## What did you change?

**Changes:**
Adds padding to tasklist. 
screen shot 2018-08-07 at 09 47 12

Fixes: 
#692 
#799 

Open task list, look at layout (accessible via dashboard->`User tasks` of a running process).

## PR-Checklist

- [ ] I've rebased the `develop` branch with my branch before finishing this PR.

### DIFF
--- a/src/modules/dashboard/dashboard.ts
+++ b/src/modules/dashboard/dashboard.ts
@@ -104,8 +104,7 @@ export class Dashboard {
   }
 
   public continueUserTask(userTaskWithProcessModel: IUserTaskWithProcessModel): void {
-    const userTask: UserTask = userTaskWithProcessModel.userTask;
-    const processModel: ProcessModelExecution.ProcessModel = userTaskWithProcessModel.processModel;
+    const {userTask, processModel} = userTaskWithProcessModel;
 
     const processModelId: string = processModel.id;
     const userTaskId: string = userTask.id;
--- a/src/modules/dashboard/dashboard.ts
+++ b/src/modules/dashboard/dashboard.ts
@@ -1,23 +1,47 @@
-import {ForbiddenError, isError, UnauthorizedError} from '@essential-projects/errors_ts';
-import {IManagementApiService, ManagementContext} from '@process-engine/management_api_contracts';
+import {
+  ForbiddenError,
+  isError,
+  NotFoundError,
+  UnauthorizedError,
+} from '@essential-projects/errors_ts';
+import {
+  IManagementApiService,
+  ManagementContext,
+  ProcessModelExecution,
+  UserTask,
+  UserTaskList,
+} from '@process-engine/management_api_contracts';
 import {inject} from 'aurelia-framework';
 import {Router} from 'aurelia-router';
 import {
   IAuthenticationService,
   NotificationType,
 } from '../../contracts/index';
+import environment from '../../environment';
 import {NotificationService} from '../notification/notification.service';
+
+interface IUserTaskWithProcessModel {
+  userTask: UserTask;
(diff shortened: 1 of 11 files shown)


LeoTT:
The problem is that currently every subsite needs to implement its own container with padding.
If we wanted to keep that system we could do:
```css
.dashboard-container  .tasklist-container {
  padding: 0;
}
```

How about the following instead?
```css
.router-view {
  position: relative;
  padding: 60px 20px;  
}

.process-def-detail {
  position: absolute;
  top: 0;
  left: 0;
  height: 100%;
  width: 100%;
}
```
That allows us to remove a couple div containers and classes and makes composition easier.
`.process-def-detail` would have to have position absolute in order to circumvent the standard padding(negative margin also possible, but more irritating imo).

Vyperus:
The subsite (like task-list)  should display itself in the space that it gets from the "container" of the site that is displaying it. It shouldn't have a padding for itself since the displaying site is repsonsible for positioning the elements.

SteffenKn:
@Vyperus The problem here is that the task list is used in the dashboard but also as an own page. For example when display user task for a specific correlation. Then the task-list has 100% width which looks disadvantageous.

cmg-dev:
Possible Solution:

Don't use the same list element twice!

Copy them and break the two lists apart.

src/modules/dashboard/dashboard.ts
```diff
@@ -1,23 +1,47 @@
-import {ForbiddenError, isError, UnauthorizedError} from '@essential-projects/errors_ts';
-import {IManagementApiService, ManagementContext} from '@process-engine/management_api_contracts';
+import {
```
Vyperus: Sort imports after known schema.

1) Framework
2) Thirdparty
3) Intern

src/modules/dashboard/dashboard.ts
```diff
@@ -95,4 +160,66 @@ export class Dashboard {
...
+    return userTasksAndProcessModels;
+  }
+
+  private async _updateUserTasks(): Promise<void> {
+    try {
+      this._userTasks = await this._getUserTasks();
+      this.succesfullRequested = true;
+    } catch (error) {
+      if (isError(error, UnauthorizedError)) {
+        this._notificationService.showNotification(NotificationType.ERROR, 'You don\'t have permission to view the task list.');
```
cmg-dev: > You do not have **the** permission (...)

src/modules/dashboard/dashboard.ts
```diff
@@ -1,23 +1,49 @@
...
+  processModel: ProcessModelExecution.ProcessModel;
+}
+
 @inject('ManagementApiClientService', 'NotificationService', 'AuthenticationService', Router)
 export class Dashboard {
 
   public showTaskList: boolean = false;
   public showProcessList: boolean = false;
+  public currentPage: number = 0;
+  public pageSize: number = 10;
```
cmg-dev: We should put this into a dashboard config object.

This should also come from the environment of aurelia.

Please create an issue and todo with this.

src/modules/dashboard/dashboard.ts
```diff
@@ -49,6 +75,45 @@ export class Dashboard {
     return true;
   }
 
+  public async activate(): Promise<void> {
+    this._getUserTasks = this._getAllUserTasks;
+    this._updateUserTasks();
+  }
+
+  public attached(): void {
+    this._getUserTasksIntervalId = window.setInterval(() => {
```
LeoTT: When is this interval cleared?

Vyperus:
@cmg-dev Done!

Issue #769, opened 2018-08-07
Solution Explorer should be opened by default


TwixRaider:
#### Describe your issue

When starting BPMN Studio for the first time it is not obvious to the user how he can open files/folder to work with. My first impulse as a new user was to search the menu and click all tab menu items.

#### How can others reproduce the issue?

Install BPMN Studio and open the app. 

#### Possible solution

It would be helpful if the solution explorer is visible by default after the start. 

#### Issue checklist

Please check the boxes in this list after submitting your Issue:

- [x] I've checked if this issue already exists
- [x] I've included all the information that i think is relevant
- [x] I've added logs and/or screenshots (if applicable)
- [x] I've mentioned PRs and issues that relate to this one

PR #773, opened 2018-08-07
✨ Open the SolutionExplorer on Studio Start

Vyperus:
## What did you change?

**Changes:**

- Makes the solution explorer show up in default when starting the app.
- Makes the solution explorer defaulting to filesystem on start on electron.

Fixes: #769 

Open BPMN-Studio on electron.

## PR-Checklist

- [x] I've rebased the `develop` branch with my branch before finishing this PR.

<details>
<summary>
<b>Merge information</b>
</summary>

Please use `:twisted_rightwards_arrows:` at the beginning of your commit title.

Example: 
<pre>
<code>
:twisted_rightwards_arrows: :bug: Fix Wrong Text Decoration at ...
</code>
</pre>

To get your commit message, just copy the first part of this pull request.

Example:
<pre>
<code>
**Changes:**

- Fixes Wrong Text Decoration at ...
- Fixes some typos
- ...

Fixes: #NumberOfFixedIssue
PR: #NumberOfThisPR
</code>
</pre>

</details>

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,6 +7,8 @@ import {NotificationService} from './modules/notification/notification.service';
 
 @inject(OpenIdConnect, 'AuthenticationService', 'NotificationService')
 export class App {
+  public showSolutionExplorer: boolean = true;
+
   private _openIdConnect: OpenIdConnect;
   private _authenticationService: AuthenticationService;
   private _router: Router;
--- a/src/app.ts
+++ b/src/app.ts
@@ -4,10 +4,10 @@ import {Router, RouterConfiguration} from 'aurelia-router';
 
 import {OpenIdConnect} from 'aurelia-open-id-connect';
 
+import {NotificationType} from './contracts/index';
 import environment from './environment';
 import {AuthenticationService} from './modules/authentication/authentication.service';
 import {NotificationService} from './modules/notification/notification.service';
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -7,6 +7,8 @@ import {NotificationService} from './modules/notification/notification.service';
 
 @inject(OpenIdConnect, 'AuthenticationService', 'NotificationService')
 export class App {
+  public showSolutionExplorer: boolean = true;
```
alexanderkasten: Also if I set this variable to false the solution explorer is still shown.
I think it would be better if we handle this variable only in the navbar because it is declared there too with a two way binding

SteffenKn: The question is:
Should the Property Panel keep the state after reload/restart of the BPMN-Studio, or if it should always open on start.

In the first case we need to fix setting the Hide/Show state to localStorage, in the second case we should remove the localStorage part completely.

cmg-dev: A1: It should keep the state.

cmg-dev: @alexanderkasten 

Handling this in the navbar is a hack.

We will need to introduce a nice cut of the components of the studio.

As this is an application state, this needs to be in the `app.ts` IMHO.

But if this is not possible by now: open an issue and do this later.

PR #1239, opened 2019-01-04
[WIP] 🐛 Fix unsaved changes Modal when Navigating back from the Diff or XML to the Detail view

### LINKED ISSUE #1236: 🐛 Save modal gets displayed when trying to use the back button on the Diff view
#### Describe your issue

When trying to use the _Back_ Button, the _Unsaved Changes_ modal gets displayed. This is unecessary, when the User opened the Diff view from the detail view since no changes will be lost. 

However its necessary if the User wants to open another diagram from the diff view.

* Open a Diagram
* Make some Changes
* Use the Diff button to go back
* See, that the modal is displayed.

#### Possible solution

Its necessary to check, if the User want's to navigate to the _Details_ or _xml_  view. If so, its not necessary to display the _Unsaved Changes_ Modal.

The problem is, that you currently can't obtain the destination route from the `aurelia-router` and therefore can't distinguish, if the user wants to navigate back to the diff or the detail view.

#### Issue checklist

robpal94:
---

**Changes:**

1. Fix a bug, where the _Unsaved Changes_ Modal was displayed, even if its not necessary. This is the case when the user tires to navigate from the _Diff_ or _XML_ - View back to the _Detail_ view. 

**Issues:**

Closes #1236 

PR: #?

---

See the [examples](#example) for inspiration.

> Describe how others can test your changes (you can remove this section for typo fixes etc.)

## PR-Checklist

- [ ] I've tested **all** changes included in this PR.
- [ ] I've also reviewed this PR myself before submitting (e.g. for scrambled letters, typos, etc.).
- [ ] I've rebased the `develop` branch with my branch before finishing this PR.
- [x] I've **summarized all changes** in a list above.
- [x] I've mentioned all **PRs, which relate to this one**.
- [x] I've prefixed my Pull Request title is according to [gitmoji guide](https://gitmoji.carloscuesta.me/).

## Example

<details>
<summary>
:warning: Before merging please click here to expand and follow the guide.
</summary>
<br>

Please use `:twisted_rightwards_arrows:` at the beginning of your merge commit title.

Example 1:

<pre>
<code>
:twisted_rightwards_arrows: :bug: Fix Wrong Text Decoration at ...
</code>
</pre>

To get your commit message, just copy the first part of this pull request.

Example 2:
<pre>
<code>
**Changes:**

- Fixes Wrong Text Decoration at ...
- Fixes some typos
- ...

**Issues:**

Closes #NumberOfFixedIssue

PR: #NumberOfThisPR
</code>
</pre>
</details>

## Emojis

<details>
<summary>
Expand for a list of most used Emojis.
</summary>
<br>

Please prefix your commit messages with an Emoji.

Ref: https://gitmoji.carloscuesta.me/

| Description              | Glyphe               | Emoji  |
|--------------------------|----------------------|--------|
| Bugfix                   | `:bug:`              | 🐛     |
| Configuration releated   | `:wrench:`           | 🔧     |
| Cosmetic                 | `:lipstick:`         | 💄     |
| Dependencies Downgrade   | `:arrow_down:`       | ⬇️     |
| Dependencies Upgrade     | `:arrow_up:`         | ⬆️     |
| Formatting               | `:art:`              | 🎨     |
| Improving Performance    | `:zap:`              | ⚡️     |
| Initial commit           | `:tada:`             | 🎉     |
| Linter                   | `:rotating_light:`   | 🚨     |
| Miscellaneous            | `:package:`          | 📦     |
| New Feature              | `:sparkles:`         | ✨     |
| Refactoring Code         | `:recycle:`          | ♻️     |
| Releasing / Version tags | `:bookmark:`         | 🔖     |
| Removing Stuff           | `:fire:`             | 🔥     |
| Tests                    | `:white_check_mark:` | ✅     |
| Work In Progress (WIP)   | `:construction:`     | 🚧     |

</details>


src/modules/design/design.ts
```diff
@@ -167,11 +189,11 @@ export class Design {
   }
 
   public async canDeactivateModal(): Promise<boolean> {
-    if (this._suppressSaveChangesModal) {
+    /*if (this._suppressSaveChangesModal) {
       this._suppressSaveChangesModal = false;
 
       return true;
-    }
+    }*/
```
cmg-dev: If this is no longer needed -> delete!

src/modules/design/design.ts
```diff
@@ -153,9 +149,8 @@ export class Design {
     this._eventAggregator.publish(environment.events.bpmnio.togglePropertyPanel);
   }
 
-  public async canDeactivate(): Promise<Redirect> {
-    const modalResult: boolean = await this.canDeactivateModal();
-
+  public async canDeactivate(destinationInstruction: NavigationInstruction): Promise<Redirect> {
+    const modalResult: boolean = await this.canDeactivateModal(destinationInstruction);
```
Vyperus: Name this variable like you would form a sentence in the if clause like:

`const userCanDeactivateRoute: boolean`

src/modules/design/design.ts
```diff
@@ -153,9 +149,8 @@ export class Design {
...
     if (!modalResult) {
```
Vyperus: Same here.

src/modules/design/design.ts
```diff
@@ -265,4 +256,39 @@ export class Design {
...
+   */
+  private _modalCanBeSuppressed(destinationInstruction: NavigationInstruction): boolean {
+    const oldView: string = this._router.currentInstruction.params.view;
+    const destinationView: string = destinationInstruction.params.view;
+
+    const navToDiffOrXmlFromDetail: boolean =
+      oldView === 'detail'
+      && (destinationView === 'xml' || destinationView === 'diff');
+
+    const navFromToDiffXml: boolean =
```
Vyperus: I understand for what you are going for here but please go for it like this:

`const navFromDiffToXml: boolean = oldView === 'diff' && destinationView === 'xml';
and
`const navFromXmlToDiff: boolean = oldView === 'xml' && destinationView === 'diff';

By simply splitting this two variables that you combined into one you can increase the readability by far.

src/modules/design/design.ts
```diff
@@ -265,4 +256,39 @@ export class Design {
...
+      (oldView === 'xml' && destinationView === 'diff')
+      || (oldView === 'diff' && destinationView === 'xml');
+
+    const navFromDiffXmlToDetail: boolean =
+      destinationView === 'detail'
```
Vyperus: Can switch those? The Check for the "oldView" first and then the "destinationView"? This way you stay consistent with your previous checks and it makes it a bit easier to read imo.

src/modules/design/design.ts
```diff
@@ -265,4 +256,39 @@ export class Design {
...
+      && (oldView === 'diff' || oldView === 'xml');
+
+    return navToDiffOrXmlFromDetail || navFromToDiffXml || navFromDiffXmlToDetail;
```
Vyperus: Here you need to add the new variables of course.

src/modules/design/design.ts
```diff
@@ -265,4 +256,44 @@ export class Design {
...
+
+    const navFromDiffXmlToDetail: boolean =
+      (oldView === 'diff' || oldView === 'xml')
+      && destinationView === 'detail';
+
+    const shouldModalBeSuppressed: boolean =
+      navToDiffOrXmlFromDetail
+      || navFromDiffToXml
+      || navFromXmlToDiff
+      || navFromDiffXmlToDetail;
```
LeoTT: How about:
```ts
const shouldModalBeSuppressed: boolean = ['xml', 'diff', 'design'].includes(oldView) 
 && ['xml', 'diff', 'design'].includes(destinationView);
```

robpal94: Then it would be possible to navigate to a routes, where the oldView and destinationView are equal. This would lead to an additional check, if the old- and Destination Diagram names are the same.

LeoTT: I don't understand why that makes a difference. When is `oldView` and `destinationView` the same and shouldn't the model be supressed either way? You could also add `&& oldView !== destinationView` if needed.

My problem with the code above is, that a very simple thing happens and it looks very complicated.
Generally I think it's often good to just write dumb rules down instead of trying to express it with a clever function but in the above code it's just not clear due to linebreaks & js-syntax and inconsistency.

If you want to write all directions explicitly, I would do something like this:
```ts
    // tslint:disable-next-line:typedef
    const navigatesFromTo = (from: string, to: string): boolean =>
      (oldView === from && to === destinationView);
    
    const shouldModalBeSuppressed: boolean = navigatesFromTo('diff', 'xml')
      || navigatesFromTo('diff', 'detail')
      || navigatesFromTo('xml', 'detail')
      || navigatesFromTo('xml', 'diff')
      || navigatesFromTo('detail', 'xml')
      || navigatesFromTo('detail', 'diff');
```

robpal94: > When is oldView and destinationView the same and shouldn't the model be supressed either way? You could also add && oldView !== destinationView if needed.

No, because when the User wants to navigate from one detail to another and has unsaved changes, the modal needs to be shown to inform the user about these.

LeoTT: Good catch.
But please make the rules consistent. If you don't like the previous suggestion, I would also take something like:
```ts
    const navFromDiffToXml: boolean = oldView === 'diff' && destinationView === 'xml';
    const navFromDiffToDetail: boolean = oldView === 'diff' && destinationView === 'detail';
    const navFromXmlToDiff: boolean = oldView === 'xml' && destinationView === 'diff';
    const navFromXmlToDetail: boolean = oldView === 'xml' && destinationView === 'detail';
... two more rules
```
Right now some rules contain an `or` and one of them goes `....To...From...` and not `...From...To` like the others.

src/modules/design/design.ts
```diff
@@ -289,4 +280,37 @@ export class Design {
...
+    const destinationView: string = destinationInstruction.params.view;
+
+    const navigatesFromTo: Function = (from: string, to: string): boolean =>
+      (oldView === from && to === destinationView);
+
+    const shouldModalBeSuppressed: boolean = navigatesFromTo('diff', 'xml')
+      || navigatesFromTo('diff', 'detail')
+      || navigatesFromTo('xml', 'detail')
+      || navigatesFromTo('xml', 'diff')
+      || navigatesFromTo('diff', 'xml')
```
LeoTT: `navigatesFromTo('diff', 'xml')` is there twice.

src/modules/design/design.ts
```diff
@@ -289,4 +280,37 @@ export class Design {
...
+   *  * diff    <-->   xml
+   *
+   * @param destinationInstruction The current router instruction which contains
+   * the destination router parameters.
+   */
+  private _modalCanBeSuppressed(destinationInstruction: NavigationInstruction): boolean {
+    const oldView: string = this._router.currentInstruction.params.view;
+    const destinationView: string = destinationInstruction.params.view;
+
+    const navigatesFromTo: Function = (from: string, to: string): boolean =>
```
LeoTT: `Function` overwrites the correct type on the right side.
Now you could do `const a: number = navigatesFromTo(null);` and the compiler would be fine with it.
If we use `Function` and `any` often enough, typescript becomes just a more noisy javascript.

The correct type is `((from: string, to: string) => boolean)`, which is painful to write/read and redundant (because the right side already gives the type). I like to disable tslint for cases like these.

Do what feels right for you.

src/modules/design/design.ts
robpal94: Hmm, does your suggested function really check both ways? 
* In one case, the value of `oldView` is `xml` and the value of `destinationView` is `diff`.
* In the other case, this is switched. 

When only calling the function for one direction (suppose the first one), the expression `(oldView === from && to === destinationView)` should only evaluate to true, if the user navigates from the xml - view to the diff view but not vise versa. Or am I wrong here?

LeoTT: Refered to this a few lines above : `const shouldModalBeSuppressed: boolean = navigatesFromTo('diff', 'xml')`.

You can make that function go 2-ways if you like and call it `navigatingBetween` or something like that. Then you'd need only 3 terms here.

src/modules/design/design.ts
SteffenKn: Why not adding a type for those functions. In this example, I would suggest simply adding `type NavigateFunction = (to: string, from: string) => void;` at the top of the file (right under the imports).

LeoTT: I like type aliases, but personally still prefer letting typeinference do its job in this case. The exact type is already right there in the same line.

PR #1352, opened 2019-02-15
✨ Use Web Components as Dynamic UI

### LINKED ISSUE #1281: ✨ Display LET With DynamicUI Modal When Continuing Tasks
#### Describe your issue

When continuing a task via the task-list or dashboard view the LET should open with the dynamicUI modal instead of a dynamicUI page.

- Run a diagram
- Go to the Dashboard in the Inspect View
- Continue a waiting task
- **Notice that the dynamicUI will be displayed instead of the LET.**

#### Issue checklist

roblen45:
**Changes:**

1. Use Web Components as Dynamic UI
2. Open Live Execution Tracker out of Task List

**Issues:**

Closes #1281

PR: #1352 

1. Checkout branch feature/dynamic_ui_test
1. run `npm install && npm run electron-start-dev`
1. Create UserTask and run the process

## PR-Checklist

- [x] I've also reviewed this PR myself before submitting (e.g. for scrambled letters, typos, etc.).
- [x] I've rebased the `develop` branch with my branch before finishing this PR.
- [x] I've **summarized all changes** in a list above.
- [x] I've mentioned all **PRs, which relate to this one**.
- [x] I've prefixed my Pull Request title is according to [gitmoji guide](https://gitmoji.carloscuesta.me/).


src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -17,14 +17,10 @@ import {
 @inject('DynamicUiService', Router, Element)
 export class DynamicUiWrapper {
 
-  public cancelButtonText: string = 'Cancel';
-  public confirmButtonText: string = 'Continue';
-  public declineButtonText: string = 'Decline';
   public onButtonClick: (action: 'cancel' | 'proceed' | 'decline') => void;
```
ElRaptorus: I'd move these strings into an enum and then use that enum as a type for the `action` parameter.

Like so:
```ts
enum ButtonClickActions {
  cancel = 'cancel',
  process = 'proceed',
  decline = 'decline',
}

public onButtonClick: (action: ButtonClickActions) => void;
```

Looks a lot cleaner imho and you'll be able to reuse that signature more effectively.

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -52,7 +48,7 @@ export class DynamicUiWrapper {
     this._activeSolutionEntry = solutionEntry;
   }
 
-  public async handleUserTaskButtonClick(action: 'cancel' | 'proceed' | 'decline'): Promise<void> {
+  public async handleUserTaskButtonClick(action: 'cancel' | 'proceed', userTask: any): Promise<void> {
```
ElRaptorus: You could use the aforementioned enum here as a type as well.

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -148,23 +95,23 @@ export class DynamicUiWrapper {
     });
   }
 
-  private _finishUserTask(action: 'cancel' | 'proceed' | 'decline'): Promise<void> {
+  private _finishUserTask(action: 'cancel' | 'proceed' | 'decline', userTask: any): Promise<void> {
```
ElRaptorus: See above. Perfect UseCase for that enum.

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
     const taskIsAnUserTask: boolean = taskWithProcessModel.userTask !== undefined;
```
ElRaptorus:
```suggestion
    const taskIsAUserTask: boolean = taskWithProcessModel.userTask !== undefined;
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
...
 
     const correlationId: string = taskIsAnUserTask
```
ElRaptorus:
```suggestion
    const correlationId: string = taskIsAUserTask
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
...
-                                ? taskWithProcessModel.userTask.correlationId
-                                : taskWithProcessModel.manualTask.correlationId;
+      ? taskWithProcessModel.userTask.correlationId
+      : taskWithProcessModel.manualTask.correlationId;
 
     const tasksProcessModelId: string = taskIsAnUserTask
```
ElRaptorus:
```suggestion
    const tasksProcessModelId: string = taskIsAUserTask
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
...
 
     const taskIsFromCallActivity: boolean = taskWithProcessModel.processModel.id !== tasksProcessModelId;
 
     const processModelId: string = taskIsFromCallActivity
-                                 ? tasksProcessModelId
-                                 : taskWithProcessModel.processModel.id;
+      ? tasksProcessModelId
+      : taskWithProcessModel.processModel.id;
 
     const taskId: string = taskIsAnUserTask
```
ElRaptorus:
```suggestion
    const taskId: string = taskIsAUserTask
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
...
-                         ? taskWithProcessModel.userTask.id
-                         : taskWithProcessModel.manualTask.id;
+      ? taskWithProcessModel.userTask.id
+      : taskWithProcessModel.manualTask.id;
 
     const processInstanceId: string = taskIsAnUserTask
```
ElRaptorus:
```suggestion
    const processInstanceId: string = taskIsAUserTask
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -142,28 +142,28 @@ export class TaskList {
...
-                                    ? taskWithProcessModel.userTask.processInstanceId
-                                    : taskWithProcessModel.manualTask.processInstanceId;
+      ? taskWithProcessModel.userTask.processInstanceId
+      : taskWithProcessModel.manualTask.processInstanceId;
```
ElRaptorus: Given the number of times this conditional assignment is repeated, it might make sense to extract that userTask/manualTask into a variable. 

Then you could write all of the property extractions like this:

```ts
const taskIsAUserTask: boolean = taskWithProcessModel.userTask !== undefined;
// Not sure which type you're using here. FlowNodeInstance, I'm guessing, based on the extracted properties? 
const taskToProcess: yourTypeHere = taskIsAUserTask 
  ? taskWithProcessModel.userTask
  : taskWithProcessModel.manualTask;

const correlationId: string = taskToProcess.correlationId;
const tasksProcessModelId: string = taskToProcess.processModelId;
const taskId: string = taskToProcess.id;
const processInstanceId: string = taskToProcess.processInstanceId;
```

src/modules/inspect/task-list/task-list.ts
```diff
@@ -289,20 +289,20 @@ export class TaskList {
       return otherCorrelation.id === correlationId;
     });
 
-    const correlationWasNotFound: boolean = correlation === undefined || correlation === null;
+    const correlationWasNotFound: boolean = correlation === undefined || correlation === null;
```
ElRaptorus: I've always preferred the shorter version of this.

```suggestion
    const correlationWasNotFound: boolean = !correlation;
```

Looks nicer, imho.
Just a suggestion, though.

index.html
```diff
@@ -1,44 +1,47 @@
...
+      // in an electron-environment, we musn't try to load umd-modules via window.module
+      if (window.module) {
+        delete window.module;
+      }
+    </script>
+    <script src="scripts/vendor-bundle.js" data-main="aurelia-bootstrapper"></script>
+    <script>
+      // this disables a bluebird-memory-leak and drastically improves performance!
+      // see https://github.com/aurelia/framework/issues/755
+      Promise.config({ longStackTraces: false });
```
cmg-dev: ist this in the right place? I suppose this will only be executed, if the index is loaded as view.

But this is not always the case, I believe.

Can you move this to a startup context?

How does this relate to backend services, where we use bluebird as well? \cc @ElRaptorus

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.html
```diff
@@ -1,20 +1,13 @@
...
+    </dynamic-usertask-component>
   </template>
   <template if.bind="isHandlingManualTask">
-    <div class="dynamic-ui__button--wrapper">
-      <button class="btn btn-default dynamic-ui__button" click.delegate="handleManualTaskButtonClick('cancel')">Cancel</button>
-      <button class="btn btn-success dynamic-ui__button" click.delegate="handleManualTaskButtonClick('proceed')">Continue</button>
-    </div>
+    <manualtask-component manualtask.bind="currentManualTask"
+      continued.delegate="handleManualTaskButtonClick('proceed')"
+      canceled.delegate="handleManualTaskButtonClick('cancel')"></manualtask-component>
```
cmg-dev:
```suggestion
      canceled.delegate="handleManualTaskButtonClick('cancel')">
    </manualtask-component>
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.html
```diff
@@ -1,20 +1,13 @@
...
   <template if.bind="isHandlingUserTask">
-    <confirm-widget if.bind="isConfirmUserTask" user-task-config.bind="currentUserTask.data"></confirm-widget>
-    <form-widget if.bind="isFormUserTask" user-task-config.bind="currentUserTask.data"></form-widget>
-    <div class="dynamic-ui__button--wrapper">
-      <button if.bind="cancelButtonText" class="btn btn-default dynamic-ui__button" click.delegate="handleUserTaskButtonClick('cancel')">${cancelButtonText}</button>
-      <button if.bind="confirmButtonText" class="btn btn-success dynamic-ui__button" click.delegate="handleUserTaskButtonClick('proceed')" id="dynamic-ui-wrapper-continue-button">${confirmButtonText}</button>
-      <button if.bind="declineButtonText" class="btn btn-danger dynamic-ui__button dynamic-ui__button--decline" click.delegate="handleUserTaskButtonClick('decline')">${declineButtonText}</button>
-    </div>
+    <dynamic-usertask-component submitted.delegate="handleUserTaskButtonClick('proceed', $event.detail)"
+      canceled.delegate="handleUserTaskButtonClick('cancel')" usertask.bind="currentUserTask">
```
cmg-dev:
```suggestion
      canceled.delegate="handleUserTaskButtonClick('cancel')"
      usertask.bind="currentUserTask">
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
cmg-dev: probably even be a little bit more specific, like DynamicUi.ButtonClickActions ?

But this would require namespaces to work.

src/modules/inspect/task-list/task-list.ts
cmg-dev: and is more robust as well.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -109,6 +110,14 @@ export class LiveExecutionTracker {
     this._parentProcessInstanceId = await this._getParentProcessInstanceId();
     this._parentProcessModelId = await this._getParentProcessModelId();
 
+    const hasTaskId: boolean = routeParameters.taskId !== undefined
+      && routeParameters.taskId !== null && routeParameters.taskId !== '';
```
cmg-dev:
```suggestion
      && routeParameters.taskId !== null
      && routeParameters.taskId !== '';
```

I would spend a line for each rule

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -109,6 +110,14 @@ export class LiveExecutionTracker {
     this._parentProcessInstanceId = await this._getParentProcessInstanceId();
     this._parentProcessModelId = await this._getParentProcessModelId();
 
+    const hasTaskId: boolean = routeParameters.taskId !== undefined
```
cmg-dev:
```suggestion
    const routeParameterContainTaskId: boolean = routeParameters.taskId !== undefined
```

I find this a much better name for the rule.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -109,6 +110,14 @@ export class LiveExecutionTracker {
...
+      && routeParameters.taskId !== null && routeParameters.taskId !== '';
+
+    if (hasTaskId) {
```
cmg-dev:
```suggestion
    if (routeParameterContainTaskId) {
```

Because on the using-side it will be much nicer to read.

index.html
ElRaptorus: Not sure. 
If this is run in a separate application context, I don't think this will cause any issue at all. 

But if it does not, it will prevent the backend from setting its own Bluebird config; which includes the enabling of Promise-Cancellation and that would be very problematic. 

It would explain those `cannot enable cancellation after promises are in use` errors that @roblen45 and @MaFa961 ran into a couple of times. 

Given that the linked issue was closed some time ago, I'd recommend to just remove that config override altogether,

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
5ebastianMeier: @cmg-dev Namespaces do work, we use them in the contracts packages already - so your proposal looks fine

index.html
Vyperus: That is actually something we took with us from... a long time ago.

I think we can safely delete this.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -245,10 +255,10 @@ export class LiveExecutionTracker {
     // Get all elements that can have a token
     const allElements: Array<IShape> = this._elementRegistry.filter((element: IShape): boolean => {
       const elementCanHaveAToken: boolean = element.type !== 'bpmn:SequenceFlow'
-                                         && element.type !== 'bpmn:Collaboration'
-                                         && element.type !== 'bpmn:Participant'
-                                         && element.type !== 'bpmn:Lane'
-                                         && element.type !== 'label';
+        && element.type !== 'bpmn:Collaboration'
```
Vyperus: I think the deleted writing method of this was far more readable.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -301,12 +311,12 @@ export class LiveExecutionTracker {
 
     const activeManualAndUserTasks: Array<IShape> = elements.filter((element: IShape) => {
       const elementIsAUserOrManualTask: boolean = element.type === 'bpmn:UserTask'
-                                               || element.type === 'bpmn:ManualTask';
+        || element.type === 'bpmn:ManualTask';
```
Vyperus: Same here.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -443,7 +453,7 @@ export class LiveExecutionTracker {
     const callActivityTarget: CorrelationProcessModel = correlation.processModels
       .find((correlationProcessModel: CorrelationProcessModel): boolean => {
         const targetProcessModelFound: boolean = correlationProcessModel.parentProcessInstanceId === this.processInstanceId
-                                              && correlationProcessModel.processModelId === callActivityTargetId;
+          && correlationProcessModel.processModelId === callActivityTargetId;
```
Vyperus: Same here.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -457,7 +467,7 @@ export class LiveExecutionTracker {
     this.selectedFlowNode = event.element;
 
     const clickedElementIsNotAUserOrManualTask: boolean = clickedElement.type !== 'bpmn:UserTask'
-                                                       && clickedElement.type !== 'bpmn:ManualTask';
+      && clickedElement.type !== 'bpmn:ManualTask';
```
Vyperus: Same here.

src/modules/task-dynamic-ui/task-dynamic-ui.html
```diff
@@ -3,10 +3,8 @@
   <require from="../dynamic-ui-wrapper/dynamic-ui-wrapper"></require>
   <div class="dynamic-ui-wrapper">
     <div class="dynamic-ui">
-      <template if.bind="userTask || manualTask">
-        <h3>${taskName}</h3>
-      </template>
-      <dynamic-ui-wrapper view-model.ref="dynamicUiWrapper" is-modal.bind="isModal" close-modal.trigger="clearTasks()"></dynamic-ui-wrapper>
+      <dynamic-ui-wrapper view-model.ref="dynamicUiWrapper" is-modal.bind="isModal" close-modal.trigger="clearTasks()">
```
Vyperus:
```suggestion
      <dynamic-ui-wrapper
        view-model.ref="dynamicUiWrapper"
        is-modal.bind="isModal"
        close-modal.trigger="clearTasks()">
```

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -781,7 +791,7 @@ export class LiveExecutionTracker {
       const couldNotGetXml: boolean = xml === undefined;
       if (couldNotGetXml) {
         const notificationMessage: string = 'XML could not be found. If the error persists, '
-                                          + 'try reopening the Live Execution Tracker or restarting the process.';
+          + 'try reopening the Live Execution Tracker or restarting the process.';
```
Vyperus: Again I think the other way it was far more readable. We should stay consistent.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -799,7 +809,7 @@ export class LiveExecutionTracker {
       const colorizingFailed: boolean = colorizedXml === undefined;
       if (colorizingFailed) {
         const notificationMessage: string = 'Could not get tokens. If the error persists, '
-                                          + 'try reopening the Live Execution Tracker or restarting the process.';
+          + 'try reopening the Live Execution Tracker or restarting the process.';
```
Vyperus: Same here.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -873,7 +883,7 @@ export class LiveExecutionTracker {
     const allActiveCorrelationsOrRequestError: Array<DataModels.Correlations.Correlation> | RequestError = await getActiveCorrelations();
 
     const couldNotGetCorrelation: boolean = allActiveCorrelationsOrRequestError === RequestError.ConnectionLost
-                                         || allActiveCorrelationsOrRequestError === RequestError.OtherError;
+      || allActiveCorrelationsOrRequestError === RequestError.OtherError;
```
Vyperus: Same here.

src/modules/inspect/task-list/task-list.ts
```diff
@@ -140,30 +140,25 @@ export class TaskList {
...
-                                      ? taskWithProcessModel.userTask.processModelId
-                                      : taskWithProcessModel.manualTask.processModelId;
+    const tasksProcessModelId: string = taskToProcess.processModelId;
 
     const taskIsFromCallActivity: boolean = taskWithProcessModel.processModel.id !== tasksProcessModelId;
 
     const processModelId: string = taskIsFromCallActivity
-                                 ? tasksProcessModelId
-                                 : taskWithProcessModel.processModel.id;
+      ? tasksProcessModelId
```
Vyperus: And here.

src/modules/inspect/task-list/task-list.ts
```diff
@@ -140,30 +140,25 @@ export class TaskList {
 
   public continueTask(taskWithProcessModel: IUserTaskWithProcessModel & IManualTaskWithProcessModel): void {
     const taskIsAnUserTask: boolean = taskWithProcessModel.userTask !== undefined;
+    const taskToProcess: DataModels.UserTasks.UserTask | DataModels.ManualTasks.ManualTask = taskIsAnUserTask
+      ? taskWithProcessModel.userTask
```
Vyperus: And here.

index.html
```diff
@@ -1,44 +1,42 @@
 <!DOCTYPE html>
 <html>
-  <head>
+
+<head>
     <meta charset="utf-8">
     <title>BPMN-Studio</title>
```
alexanderkasten: Please fix the indentation of the file

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -52,36 +51,20 @@ export class DynamicUiWrapper {
     this._activeSolutionEntry = solutionEntry;
   }
 
-  public async handleUserTaskButtonClick(action: 'cancel' | 'proceed' | 'decline'): Promise<void> {
-    const actionCanceled: boolean = action === 'cancel';
+  public async handleUserTaskButtonClick(action: ButtonClickActions, userTask: any): Promise<void> {
```
alexanderkasten: Why is userTask any here?
I would expect that 
```suggestion
  public async handleUserTaskButtonClick(action: ButtonClickActions, userTask: DataModels.UserTasks.UserTask): Promise<void> {
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -148,23 +98,23 @@ export class DynamicUiWrapper {
     });
   }
 
-  private _finishUserTask(action: 'cancel' | 'proceed' | 'decline'): Promise<void> {
+  private _finishUserTask(action: ButtonClickActions, userTask: any): Promise<void> {
```
alexanderkasten: same here
```suggestion
  private _finishUserTask(action: ButtonClickActions, userTask: DataModels.UserTasks.UserTask): Promise<void> {
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -148,23 +98,23 @@ export class DynamicUiWrapper {
...
     }
 
-    const correlationId: string = this.currentUserTask.correlationId;
-    const processInstanceId: string = this.currentUserTask.processInstanceId;
-    const userTaskInstanceId: string = this.currentUserTask.flowNodeInstanceId;
-    const userTaskResult: DataModels.UserTasks.UserTaskResult = this._getUserTaskResults();
+    const correlationId: string = userTask.correlationId;
+    const processInstanceId: string = userTask.processInstanceId;
+    const userTaskInstanceId: string = userTask.userTaskInstanceId;
+    const userTaskResult: DataModels.UserTasks.UserTaskResult = userTask.results;
```
LeoTT:
```ts
const {correlationId, processInstanceId, userTaskInstanceId, results} = userTask;
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -148,23 +98,23 @@ export class DynamicUiWrapper {
...
 
     this._dynamicUiService.finishUserTask(this._identity,
-                                          processInstanceId,
-                                          correlationId,
-                                          userTaskInstanceId,
-                                          userTaskResult);
+      processInstanceId,
+      correlationId,
+      userTaskInstanceId,
+      userTaskResult);
```
LeoTT: Previous indentation was correct according to our [styleguide](https://github.com/5minds/typescript-guidelines/tree/develop/tslint-config-5minds)(4.1).

Also the way the tsserver indents my code by default, I think.

src/modules/inspect/task-list/task-list.ts
```diff
@@ -140,30 +140,25 @@ export class TaskList {
...
 
-    const taskId: string = taskIsAnUserTask
-                         ? taskWithProcessModel.userTask.id
-                         : taskWithProcessModel.manualTask.id;
+    const taskId: string = taskToProcess.id;
 
-    const processInstanceId: string = taskIsAnUserTask
-                                    ? taskWithProcessModel.userTask.processInstanceId
-                                    : taskWithProcessModel.manualTask.processInstanceId;
+    const processInstanceId: string = taskToProcess.processInstanceId;
```
LeoTT: Could use object destructuring for the assignments above as well.

src/modules/inspect/task-list/task-list.ts
```diff
@@ -289,20 +284,20 @@ export class TaskList {
       return otherCorrelation.id === correlationId;
     });
 
-    const correlationWasNotFound: boolean = correlation === undefined || correlation === null;
+    const correlationWasNotFound: boolean = !correlation;
```
LeoTT: The explicit statement is better imo.

Consider:
```ts
const numberIsNotSet: boolean = !num;
```
Looks right, until `num` is `0`. That would make the negation also true. 


That being said, the real mistake was the null-check. `find` can only return `undefined`(or the correlation).
Therefore `const correlationWasNotFound: boolean = correlation === undefined` does the job.

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -33,7 +33,8 @@ type RouteParameters = {
   diagramName: string,
   solutionUri: string,
   correlationId: string,
-  processInstanceId: string;
+  processInstanceId: string,
+  taskId: string,
```
LeoTT: I would recommend to include `null` and `undefined` if that type is expected. That way, the user of the type knows not to trust it having a value.

`   taskId: string | undefined | null,`

That looks stupid, of course, but the real problem is that taskId could be either of those values and we don't know which.

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -186,35 +136,15 @@ export class DynamicUiWrapper {
     const manualTaskInstanceId: string = this.currentManualTask.flowNodeInstanceId;
 
     this._dynamicUiService.finishManualTask(this._identity,
-                                            processInstanceId,
-                                            correlationId,
-                                            manualTaskInstanceId);
+      processInstanceId,
+      correlationId,
+      manualTaskInstanceId);
```
LeoTT: See above, previous indentation was correct (according to styleguide).

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -7,24 +7,23 @@ import {IIdentity} from '@essential-projects/iam_contracts';
...
 
-  public cancelButtonText: string = 'Cancel';
-  public confirmButtonText: string = 'Continue';
-  public declineButtonText: string = 'Decline';
-  public onButtonClick: (action: 'cancel' | 'proceed' | 'decline') => void;
-  @bindable({changeHandler: 'userTaskChanged'}) public currentUserTask: DataModels.UserTasks.UserTask;
-  @bindable({changeHandler: 'manualTaskChanged'}) public currentManualTask: DataModels.ManualTasks.ManualTask;
-  @bindable() public isConfirmUserTask: boolean = false;
-  @bindable() public isFormUserTask: boolean = false;
+  public onButtonClick: (action: ButtonClickActions) => void;
```
LeoTT: So the pattern here is, that the DynamicUiWrapper contains a TaskDynamicUi, which will set this function at some point with: 
```ts
    this.dynamicUiWrapper.onButtonClick = (action: string): void => {
      this._finishTask(action);
    };
```
Correct me if I'm wrong, that was my understanding.

I find that callback-pattern confusing and it has the downside that onButtonClick is not set at all times.
I would recommend using Events instead.

If this must stay, I would add the `undefined` type.

```ts
   public onButtonClick: ((action: ButtonClickActions) => void) | undefined;
```

src/modules/live-execution-tracker/live-execution-tracker.ts
ElRaptorus: I think the question here is if it should at all be possible to have an `undefined` taskId. 
We're talking about the `id` here, not the `name`. 
If that ID is missing, for whatever reason, I think we have a greater problem here, because we need that ID for pretty much every query we have.


I'd leave this definition as it is.

src/modules/inspect/task-list/task-list.ts
ElRaptorus: Usually this is a fair point, but I wonder if `0` should really constitute a valid ID for a correlation?

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
ElRaptorus: I don't think adding `| undefined` would be helpful.
Any object can at any time be undefined, regardless of its type. But we don't include undefined in the type signatures. Why would we?
If we did, we'd have to write all typedefs like so:

```ts
  @bindable() public currentUserTask: DataModels.UserTasks.UserTask | undefined;
  @bindable() public currentManualTask: DataModels.ManualTasks.ManualTask | undefined;
``` 

Because these objects could also be undefined at some point.

I think just leaving the typedef as it is and performing a check for "undefined", prior to using it, suffices.
Such a check is performed here, for example:
https://github.com/process-engine/bpmn-studio/blob/6a275aaed22217e5c2df4f52d9a5f3678cf2e6e3/src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts#L121

But I understand your confusion.
Perhaps an explanatory comment on top of this callback might clear this up a bit.

Dunno, maybe something like 
```suggestion
  /**
   * This callback is set by the TaskDynamicUi.
   */
  public onButtonClick: (action: ButtonClickActions) => void;
```

src/modules/inspect/task-list/task-list.ts
LeoTT: No, the example has no relation to this code. Was just trying to show one case where js negation can trick you.

Does not happen too often that negation does not do what you want, but when it happens it can really boggle your mind. I had that bug once and it took me way too long to get it. Therefore I would make a habit of doing the explicit thing with `=== undefined`

ElRaptorus: Ok, fair point.

src/modules/live-execution-tracker/live-execution-tracker.ts
LeoTT: There is a check for `undefined` and `null` later on in this file.
I made that comment because if a variable has a type string, I would like to be able to trust it being a string and not `null` or `undefined`. 
If the id is always set, this should be `string` and the check is unnecessary.

ElRaptorus: I see. Didn't know about that null-check. 
I find that rather confusing, perhaps @SteffenKn can shed some light on this?

If that parameter really is optional, then I'd agree that it should be marked as such like `taskId?: string`.
Goes for the other parameters as well of course.

roblen45:
> I don't know why but I can not fill text to all formfields 🤔
> 
> feb-18-2019 09-59-42
> 
> I also get this error. Maybe it's part of form field error?
> 
> bildschirmfoto 2019-02-18 um 10 01 59
> 
> Edit:
> 
> That only happens if I set the form field type to integer or decimal

You can only insert digits in a number or long form field.

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
LeoTT: I realize that we would have to use `| undefined` in many cases and I honestly think that would be good.

If possible it's a good thing to initialize variables. A variable not being set immediatly should be the special case. 
TypeScript does not really help you to avoid all the `cannot read property _ of undefined` errors and all the `|| thing === undefined || thing === null`-checks unless you differentiate between types which are always set and types which are maybe set.
If something is `undefined` at runtime it should have a reason for that. In case of the binding examples you gave I think it does make sense. In the constructor you cannot access those properties, so the type `DataModels.UserTasks.UserTask` is not reliable.
You could also make `type Maybe<a> = a | undefined;`. Looks a bit like haskell then.

I don't want to add that type everywhere, I want to have the variables initialized with the right value if possible(and eliminate null).

I realize now that my opinion may be opinionated ;)
Just wanted to clarify my motivation.

Vyperus:
So I added the requested Changes i was able to identify in this discussion.

Before merging however I want to wait for some some BugFixes in https://github.com/process-engine/dynamic_ui_core.

@LeoTT @ElRaptorus 
Can you check your requested changes please?

src/modules/inspect/task-list/task-list.ts
LeoTT:
```suggestion
    const correlationWasNotFound: boolean = correlation === undefined;
```

src/modules/live-execution-tracker/live-execution-tracker.ts
```diff
@@ -109,6 +110,15 @@ export class LiveExecutionTracker {
     this._parentProcessInstanceId = await this._getParentProcessInstanceId();
     this._parentProcessModelId = await this._getParentProcessModelId();
 
+    const routeParameterContainTaskId: boolean = routeParameters.taskId !== undefined
```
LeoTT: Is the taskId really optional? 
The type suggests otherwise (https://github.com/process-engine/bpmn-studio/pull/1352/files#r257933773).

LeoTT:
@Vyperus here are the remaining suggestions:
- [object destructuring](https://github.com/process-engine/bpmn-studio/pull/1352/files#r257929982)
- [explicit undefined check](https://github.com/process-engine/bpmn-studio/pull/1352/files#r258397283)
- [check for null/undefined, where type says its not optional](https://github.com/process-engine/bpmn-studio/pull/1352/files#r258398511)

src/modules/live-execution-tracker/live-execution-tracker.ts
SteffenKn: I did not add this but it looks like the parameter is optional, so using `taskId?: string` would be correct here as @ElRaptorus said.

src/modules/live-execution-tracker/live-execution-tracker.ts
SteffenKn: The type is wrong. The taskId has to be optional, since it is only used when continuing processes.

alexanderkasten:
@Vyperus 

> This error has nothing to do with this PR

Why? IMO the dynamic ui should can handle the configured usertask types 🤔

Vyperus:
@alexanderkasten 

1) I was not able to reproduce this error.
2) The error doesn't seem like anything related to the web components.
2) See @roblen45 comment.

Can you now reproduce this error? If yes maybe these have something to do with it:
https://github.com/process-engine/dynamic_ui_core/issues/12
https://github.com/process-engine/dynamic_ui_core/issues/11
https://github.com/process-engine/dynamic_ui_core/pull/10

As I stated I wanted to wait fir this fixes before merging this anyway.

index.html
alexanderkasten: not fixed yet

index.html
```diff
@@ -1,44 +1,42 @@
...
     <link rel="icon" type="image/x-icon" href="favicon.ico">
 
     <script>
-      // if we're in the browser, we need to add <base href="/" /> to enable push state
-      if (!window.require) {
-        var base = document.createElement('base');
-        base.href = "/";
-        document.getElementsByTagName('head')[0].appendChild(base);
-      }
+        // if we're in the browser, we need to add <base href="/" /> to enable push state
```
alexanderkasten: before the indentation was correct, please undo

index.html
```diff
@@ -1,44 +1,42 @@
...
+        if (!window.require) {
+            var base = document.createElement('base');
+            base.href = "/";
```
alexanderkasten: before the indentation was correct, please undo

index.html
```diff
@@ -1,44 +1,42 @@
...
     <div aurelia-app="main" style="height: 100%;">
-      <script>
-        // see https://electron.atom.io/docs/faq/#i-can-not-use-jqueryrequirejsmeteorangularjs-in-electron
-        if (window.require) {
-          window.nodeRequire = require;
-          delete window.require;
-        }
+        <script>
+            // see https://electron.atom.io/docs/faq/#i-can-not-use-jqueryrequirejsmeteorangularjs-in-electron
+            if (window.require) {
```
alexanderkasten: before the indentation was correct, please undo

index.html
```diff
@@ -1,44 +1,42 @@
...
+                window.nodeRequire = require;
+                delete window.require;
```
alexanderkasten: before the indentation was correct, please undo

index.html
```diff
@@ -1,44 +1,42 @@
...
-      </script>
+            // in an electron-environment, we musn't try to load umd-modules via window.module
+            if (window.module) {
+                delete window.module;
+            }
+        </script>
+        <script src="scripts/vendor-bundle.js" data-main="aurelia-bootstrapper"></script>
     </div>
-  </body>
+</body>
```
alexanderkasten: before the indentation was correct, please undo

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
```diff
@@ -7,24 +7,23 @@ import {IIdentity} from '@essential-projects/iam_contracts';
...
-  IBooleanFormField,
   IDynamicUiService,
-  IEnumFormField,
   ISolutionEntry,
-  IStringFormField,
 } from '../../contracts';
 
+enum ButtonClickActions {
+  cancel = 'cancel',
+  process = 'proceed',
```
alexanderkasten:
```suggestion
  proceed = 'proceed',
```

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
alexanderkasten: still so

src/modules/dynamic-ui-wrapper/dynamic-ui-wrapper.ts
alexanderkasten: still so

package.json
```diff
@@ -126,6 +128,7 @@
     "highlightjs-line-numbers.js": "^2.2.0",
     "minimist": "^1.2.0",
     "node-fetch": "2.3.0",
+    "node-uuid": "1.4.8",
```
alexanderkasten: `node-uuid` is not by the bpmn-studio so the package who need it should bring it by himself.
That means that you should add node-uuid to the dynamic_ui_core package dependency section in its `package.json`.

Vyperus: I was confused by this since this haunts us for a while now, so I debugged this.
This is whatI came up with.

1) I deleted `node-uuid` from the package.json and aurelia.json. Here is the log from the Studio on start:
Bildschirmfoto 2019-03-09 um 10 56 51

So actually it is the `management_api_client` that need `node-uuid` but doesn't have it in the packa.json. See here:
https://github.com/process-engine/management_api_client/blob/develop/package.json
And here:
https://github.com/process-engine/management_api_client/blob/develop/src/accessors/external_accessor.ts

I wrote an Issue for this here:
https://github.com/process-engine/process_engine_runtime/issues/288

May @ElRaptorus give us some insight on this :)

For this PR:
`node-uuid` doesn't belong here but for some reason we need to start the Studio.

This may not be the answer to all of this but it may solve this for us.
Anyway this should stay here until the problem in `management_api_client` is solved and will then be handled in another PR.

ElRaptorus: @Vyperus 

As I stated [here](https://github.com/process-engine/process_engine_runtime/issues/288), the client already contains `node-uuid` as a dependency.

See here:
https://github.com/process-engine/management_api_client/blob/develop/package.json#L26

Maybe you should check the version you are using.

ElRaptorus: @Vyperus Admittedly, my knowledge about aurelia is somewhat limited, but isn't the error kinda expected, if you remove the `node-uuid` package from the aurelia.json? 🤔 
If I understand correctly, the package won't be included in the bundling any longer, if it isn't listed there.

PR #1492, opened 2019-06-06
✨ Add FlowNodeIds to LogViewer

### LINKED ISSUE #1491: ✨ Add FlowNodeId and FlowNodeInstanceId to LogViewer
## Description

It would be very helpful, if the LogViewer would contain the FlowNodeId and FlowNodeInstanceID for each LogEntry. 
Right now, one can only guess which log belongs to which FlowNode, or which FlowNodeInstance.

The values are already stored in the LogEntry data model. So adding those to the LogViewer can be done easily.

ElRaptorus:
## Changes

1. Add the columns `FlowNodeId` and `FlowNodeInstanceId` to the LogViewer.
2. Enhance the LogViewer's sorting, so that it can deal with `undefined` values.

## Issues

Closes #1491

PR: #1492

## How to test the changes

- Run a ProcessModel
- After it is finished, go to the `Inspect` view and select the logs for the executed ProcessInstance
- See that the logs will now contain the `FlowNodeId` and `FlowNodeInstanceId` for each entry.
    - Note that ProcessInstance logs will not have any of those IDs associated with them.

The view should look like this:

Bildschirmfoto 2019-06-06 um 14 15 58

### DIFF
--- a/src/contracts/inspect-correlation/inspect-panel/log-viewer/constants.ts
+++ b/src/contracts/inspect-correlation/inspect-panel/log-viewer/constants.ts
@@ -1,5 +1,7 @@
 export enum LogSortProperty {
-  Time = 'timestamp',
-  Message = 'message',
+  FlowNodeId = 'flowNodeId',
+  FlowNodeInstanceId = 'flowNodeInstanceId',
   LogLevel = 'logLevel',
+  Message = 'message',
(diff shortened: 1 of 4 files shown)


ElRaptorus:
@SteffenKn Good idea. 👍

PR #1538, opened 2019-07-04
✨ Add Remote Solution History

SteffenKn:
## Changes

1. Add Remote Solution History

PR: #1538

## How to test the changes

- Add some remote solutions
- **Notice that there will be a remote solution history, as soon as the first remote solution was opened**
- Click on any entry of the remote solution history
- **Notice that the uri remote solution will be written in the input automatically**
- **Notice that there is an online status of the process engine**


src/modules/solution-explorer/solution-explorer-panel/solution-explorer-panel.scss
```diff
@@ -57,3 +57,57 @@
 .uri-input--invalid:focus {
   border: 2px red solid;
 }
+
+
```
alexanderkasten:
```suggestion
```

src/modules/solution-explorer/solution-explorer-panel/solution-explorer-panel.scss
```diff
@@ -57,3 +57,57 @@
...
+
```
alexanderkasten:
```suggestion
```

src/modules/solution-explorer/solution-explorer-panel/solution-explorer-panel.scss
```diff
@@ -57,3 +57,57 @@
...
+
```
alexanderkasten:
```suggestion
```

PR #1719, opened 2019-09-05
✨Add expose functionality module

alexanderkasten:
## Changes

1. Add a module which exports two functions. One function to expose functions for the test client and one function to call the exposed function with the test client.
2. Fix the test (paths)

## Issues

PR: #1719 

## How to test the changes

- check whether the open a solution and save diagram tests are working

### DIFF
--- a/src/services/expose-functionality-module/expose-functionality-module.ts
+++ b/src/services/expose-functionality-module/expose-functionality-module.ts
@@ -1,0 +1,18 @@
+/* eslint-disable no-underscore-dangle */
+export function exposeFunctionForTesting(functionName: string, functionCallback: Function): void {
+  const danngerouslyInvokeObjectIsUndefined = (window as any).__dangerouslyInvoke === undefined;
+
+  if (danngerouslyInvokeObjectIsUndefined) {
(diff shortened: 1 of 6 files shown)


src/services/expose-functionality-module/expose-functionality-module.ts
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-underscore-dangle */
+export function exposeFunctionForTesting(functionName: string, functionCallback: Function): void {
+  const danngerouslyInvokeObjectIsUndefined = (window as any).__dangerouslyInvoke === undefined;
```
SteffenKn:
```suggestion
  const dangerouslyInvokeObjectIsUndefined = (window as any).__dangerouslyInvoke === undefined;
```

src/services/expose-functionality-module/expose-functionality-module.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (danngerouslyInvokeObjectIsUndefined) {
```
SteffenKn:
```suggestion
  if (dangerouslyInvokeObjectIsUndefined) {
```

PR #1741, opened 2019-09-11
♻️ Improve inspect correlation performance

alexanderkasten:
## Changes

Opening the Inspect-Correlation view with a ProcessModel which contains ca. `10k` Correlations/Instances:

before
`54484 msTotal`
after
`18690 msTotal`

Opening the view with a ProcessModel which contains `6` Correlations/Instances but with the same DB:

before
`15582 msTotal`
after
`353 msTotal`

1. Use getCorrelationsByProcessModelId instead of getAllCorrelations
2. Refactor correlation-list methods
3. Remove number column and move correlation id column to the left
4. Fix bug where the loading spinner spins forever if no correlations found
5. Always show the inspect navigation buttons on the inspect view
6. Allow opening inspect-correlation and heatmap without an active diagram or with a file system diagram

## Issues

Part of #1594 

PR: #NumberOfThisPR

## How to test the changes

- checkout the branch
- use the Inspect-correlation view with a database which contains a lot of correlations or instances
- realize that the performance is better than before


src/modules/inspect/heatmap/heatmap.scss
```diff
@@ -61,6 +63,9 @@
 }
 
 .heatmap__empty-message {
-  margin: 20px;
+  width: 100%;
+  justify-content: center;
```
alexanderkasten:
```suggestion
```

PR #1895, opened 2019-12-09
💄Use Same Cursor Style for Icon Buttons

### LINKED ISSUE #1893: 💄 Different Cursor Style for Icon Buttons
## Description

Currently we have two different cursor styles for icon buttons in the BPMN Studio.
Sometimes we use the default cursor and sometimes the pointer cursor.

## Reproduction

Hover over some icon buttons.

## Possible solution

Imho 'cursor: pointer;' should be used for every icon button.

SteffenKn:
## Changes

1. Use Same Cursor Style for Icon Buttons

## Issues

Closes #1893 

PR: #1895

## How to test the changes

- Hover over some icon buttons
- **Notice that the cursor should always look like this: cursor-pointer**

> For some reason, it is impossible to specify a cursor style for buttons in the navigation bar.
> This is caused by moving the electron title bar into the  BPMN Studio.

### DIFF
--- a/src/modules/status-bar/status-bar.html
+++ b/src/modules/status-bar/status-bar.html
@@ -5,7 +5,7 @@
   -->
   <div class="status-bar status-bar--system-macos" id="statusBarContainer">
     <div class="status-bar__left-bar" id="statusBarLeft">
-      <a if.bind="updateAvailable" class="update-button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      <a if.bind="true || updateAvailable" class="status-bar__button update-button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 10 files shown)


src/modules/status-bar/status-bar.html
```diff
@@ -5,15 +5,15 @@
   -->
   <div class="status-bar status-bar--system-macos" id="statusBarContainer">
     <div class="status-bar__left-bar" id="statusBarLeft">
-      <a if.bind="updateAvailable" class="update-button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      <a if.bind="true || updateAvailable" class="status-bar__button update-button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
ElRaptorus: Leftover?

```suggestion
      <a if.bind="updateAvailable" class="status-bar__button update-button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```

SteffenKn: 👍

PR #1944, opened 2020-01-20
🐛 Prevent IDs That Contain Spaces

SteffenKn:
## Changes

1. Prevent IDs That Contain Spaces

## Issues

Related #1943 

PR: #1944

## How to test the changes

- Open any diagram
- Select any element
- Try to add spaces to the element id


src/modules/design/property-panel/indextabs/general/sections/basics/basics.ts
```diff
@@ -264,6 +264,8 @@ export class BasicsSection implements ISection {
       .then()
       .satisfies((id: string) => this.formIdIsUnique(id) && this.isProcessIdUnique(id) && this.isDefinitionIdUnique(id))
       .withMessage('ID already exists.')
+      .satisfies((id: string) => !id.includes(' '))
```
Paulomart: you should order your validation rules by "heavyness": first all simple rules, then more complex rules.

for me it looks like the `no space`-rule looks simpler then the `isDefinitionIdUnique`-rule, therefore it should be before that

